PR #38, opened 2015-10-29
dm-check-headers updates

pipitone:
This update includes the following:

- removed blacklist and checklist functions
- now called with the top level exams dir (e.g. data/dcm) and not a list
  of exam directories
- log files are named `dm-check-headers-<exam>.log`
- only runtime errors are printed to stdout
- refactored significantly to allow testing
- added tests

Let me know what you think. Merging this will require updating all of the `run.sh` scripts to call this script properly (since the arguments have changed), and it also means that until the QC document includes the header differences, there will be no warnings printed to `run.sh` on header mismatches. (I could add a warning to stdout to this PR which we could remove later.)

### DIFF
--- a/bin/dm-check-headers.py
+++ b/bin/dm-check-headers.py
@@ -3,42 +3,38 @@
 Diffs the relevant fields of the header to determine if anything has changed in
 the protocol.
 
-Usage: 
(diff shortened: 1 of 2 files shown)


pipitone:
Superceded by #39. Closing

PR #50, opened 2016-06-15
Blacklist issues

DESm1th:
The new generated run scripts always set the --blacklist option for xnat-extract.py without checking if a blacklist.csv file exists, which may throw errors when xnat-extract.py tries to parse the table. The changes should handle this situation. 

The issues with xnat-extract.py failing to respect blacklists doesn't seem to be due to a bug, but is because some run.sh scripts don't set the blacklist option (for example the ANDT archive) so it's using the default of an empty list.

### DIFF
--- a/assets/blacklist-rm.sh
+++ b/assets/blacklist-rm.sh
@@ -2,23 +2,41 @@
 #
 # Finds and removes anything listed on the given blacklist.
 #
-# Author: Dawn E.A. Smith     Email: Dawn.Smith@camh.ca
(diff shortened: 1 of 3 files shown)


DESm1th:
Ah shoot, I forgot to mention: When my last pull request was merged an old (broken?) version of tigr-write-runsh.py and the redux version of it were added to the master branch.

pipitone:
> Ah shoot, I forgot to mention: When my last pull request was merged an
> old (broken?) version of tigr-write-runsh.py and the redux version of it
> were added to the master branch.

No worries. Want to submit a PR to reverse that change? ;-)

PR #54, opened 2016-08-08
Development

DESm1th:
There's conflict between the version of proc-freesurfer already there and mine because I moved so much around to make things easily unit tested, but I manually added in the changes from Erin's last two commits to this file so nothing will be lost if it replaces the old one.


pipitone:
Nice!!!

Unit tests, holy smokes.

DESm1th:
Yeah, I accidentally left three important lines commented out though so I had to close and submit another request! Sorry about that.

pipitone:
Hey Dawn, check out "git rebase" for a magical way of cleaning up these conflicts before you send a PR. You can also do a merge into your branch before sending the PR.

DESm1th:
I'll look into that, thank you! I had tried merging into my branch, but the current datman copy of proc-freesurfer was so different from mine after the rewrites that it screwed up my code. I ended up just manually entering Erin's last few changes, since it was just two or three small line changes (one if which I'd independently made the same bug fix to as well).

PR #124, opened 2018-05-08
 Create symlinks from RESOURCES to nii for CU1 site in OPT

josephmje:
json files still need to be created.

I'd like to update this later so it's more general and integrate with dm_link_sprl.py


DESm1th:
Hey MJ, should I wait for that update or do you mean you want to make an additional pull request later on?

DESm1th:
CU1 for OPT did have trouble because they include niftis in their zip files. This is why we do the very weird thing where we download from their server, deposit the download files in the zips folder, automatically rename them (basically just capitalize the names from their server) and then upload them to ours with dm_xnat_upload.py. I finished working out the kinks for this earlier this week, so everything should be there now but they definitely were having trouble up until this Monday. (And please let me know if you ever notice anything that's still missing from their data! I was pretty careful but their stuff has caused a lot of unforeseen issues).

josephmje:
This is ready for review now @DESm1th

josephmje:
Thanks for the feedback Dawn. I fixed comments 1, 2 and 5. 

I made the site and session arguments mutually exclusive through docopt. You can either process all the sessions from a single site (site argument) or process specific sessions (repeating session arguments). 

Also, I used datman.scanid to parse the dicom. But I agree, there needs to be a better way of getting the series number from the nifti.

josephmje:
I think I'm finally done editing.

PR #136, opened 2018-06-25
Link redcap comments and unsynced changes to symlink script

josephmje:


josephmje:
I think the symlink script was outdated in my master branch.

PR #140, opened 2018-07-05
fixed queue killer

jerdra:
i have no words


jerdra:
should be okay now, tested outputs to make sure it was correct :+1:

PR #155, opened 2018-08-15
DTIPREP from local --> scc

jerdra:


jerdra:
Made changes as requested. 
Sorry for the cringey global vars but it had to be done :cry:

PR #157, opened 2018-08-17
omit empty files in RESOURCES from resource data exists checker

josephmje:


josephmje:
i added this line: local_resources_mod = [item for item in local_resources if zf.read(item)]

josephmje:
fixed the indent

PR #158, opened 2018-08-20
Display scanlength on static qc pages for all scan types

josephmje:


jerdra:
wouldn't changing from [3] to [-1] mean we obtain the primary slice direction axis for non-fmri scans? What do we mean from scanlength? Temporal or # of slices?

josephmje:
For 4D scans, we'd get temporal. For everything else, we'd get the number of slices.

jerdra:
Number of slices might still be relevant to capture for fmri

jerdra:
also I'm not entirely sure that 3rd dimension is necessarily primary slice direction. If i'm right, Nifti convention is RAS, the S/I axis isn't always the slice encoding direction. Lmk if I'm just wrong about this

jerdra:
Sorry for spam, but see: 

http://nipy.org/nibabel/reference/nibabel.nifti1.html 
Check get_dim_info(), that'll get you the slice direction!

josephmje:
this would be running get_data_shape(), not get_dim_info()

PR #159, opened 2018-08-20
dm_bids_app to output to common directory (better fmriprep/ciftify handling).

jerdra:
with updated tests :)


bin/dm_bids_app.py
```diff
@@ -76,12 +79,13 @@ def get_bids_name(subject):
...
     '''
     
+    #Take into account that we may either start with a string (repeated site ID) or have number  
     try: 
-        sub_num = subject.split('_')[2] 
-    except IndexError: 
-        logger.error('Subject {}, invalid subject name!'.format(subject))
-        logger.error('Subject should have STUDY_SITE_SUB#_... format, exiting...')
-        raise
+        int(subject.split('_')[2][0])
```
josephmje: I was going to suggest using the Identifier class in `scan.py` from datman but is it done this way because of how OPT names their subjects?

jerdra: Good point, I was thinking of using ident since it abstracts naming specifications but I had two problems: 

One being your point on OPT naming, sometimes OPT sites use <STUDY>_<SITE>_<SITE><NUM>_<SESSION> . Ident doesn't handle these

Second being that would be the only cases where I needed datman.scan functionality, so I decided to avoid importing and dealing with handling the Scan class. 

But i see your point in that abstracting away naming details is probably better long-term. Maybe we could incorporate get_bids_name into datman.scan? 

Let me know what you guys think!

jerdra: ok Dawn agrees with implementing get_bids_name into datman.Scan. I'll do that and update my code!

josephmje: I don't see any other changes.

jerdra:
@josephmje @gabiherman updated some things to better handle scan ids for bids!

josephmje:
Everything looks ok to me! Was the new `get_numeric` function meant to be used?

jerdra:
urrghh thanks for catching that, removed it

PR #160, opened 2018-08-21
add ciftify_fmriprep module to dm_bids_app

jerdra:
Added ciftify singularity pipeline to dm_bids_app


jerdra:
@edickie I followed the formatting from your script: 
https://github.com/edickie/bids-on-scinet/blob/master/running_ciftify_container_for_camh_scc.md

Specifically: 
`SIDlist=`cd ${bids_dir}; ls -1d sub* | sed 's/sub-//g'``
Which removed the 'sub-' prefix when inputting arguments into the singularity call

Let me know if leaving 'sub-' in will work with ciftify-fmriprep, in that case i could just use $SUB and it should work out!

josephmje:
Does this replace `dm_proc_fs2hcp.py`?

jerdra:
@josephmje theoretically I think it should since both are for the purpose of converting to CIFTI format. @edickie could confirm this for sure.

josephmje:
Thanks @jerdra. Another question I had is do we care about running the group commands for mriqc and ciftify?

jerdra:
@josephmje yes definitely. 

I restricted the pipelines to participant level (+ longitudinal) since the primary application of the script is to run subjects as they come in through our nightly pipelines on the scc to maximize throughput.  

With group level summary analyses, as far as i know, they should be using the outputs from the individual runs to perform the aggregate analysis. In that case, it makes sense to have a separate script to handle that functionality using outputs (standard BIDS-app outputs) from this script.

edickie:
For the sub argument.. all bids apps take the participant id without the
"sub-" by default. I don't know if fmriprep or mriqc will strip the "sub-"
if present..

As for the "group" mode, yes, I think we need them to run for not mriqc and
ciftify.

Note: the ciftify group mode takes only seconds to run and does not need to
be submitted.

jerdra:
@edickie great! Both fmriprep and mriqc work with 'sub-' as that's how I've been doing them. I wasn't aware that you could submit ID's without 'sub-'.

Since that's possible I'll just modify $SUB globally to snip off the prefix!

jerdra:
updated changes wrt to prefix across all pipelines for consistency!

PR #167, opened 2018-09-10
dm_to_bids fix for EMP, dm_bids_app rewrite update + default logging

jerdra:
1) dm_to_bids --> added EMP to task list (long term we should push this to config so we don't edit code) 
2) dm_bids_app: 
    a) Always create .log files in <OUTPUT>/bids_logs/<APP_NAME>/<SUB>_<APP_NAME>.log
    b) Use aforementioned log outputs to handle rewrite option (search for 'Error' term)

I removed whitespace (eeeeeeeep)


jerdra:
gen_log_redirect was updated
filter_subjects was updated 
Both in dm_bids_app so you don't waste your time looking at whitespace :)

bin/dm_bids_app.py
```diff
@@ -75,137 +80,154 @@
...
     '''
-    Filter out subjects that have alrady been previously run through the BIDS-app pipeline 
+    Filters out subjects that have successfully completed the BIDS-pipeline
+    Utilizes default log output (always enabled)
 
-    Arguments: 
+    Arguments:
         subjects                List of candidate subjects to be processed through pipeline
         out_dir                 Base directory for where BIDS-app will output
+        bids_app                Name of BIDS-app (all upper-case convention)
```
DESm1th: Not required but if a non-uppercase string will cause any problems it might be wise to just use upper() to guarantee it's what you expect.

PR #168, opened 2018-09-11
Fix dm-get-session-info.py

josephmje:


josephmje:
No I want `os.path.basename()` but I can change the variable to `session_name`. That's what I really want.

PR #170, opened 2018-09-19
Multiecho mod

josephmje:


josephmje:
NOTE: All config files need to be changed before pulling this mod into our filesystem.

josephmje:
Submitted a pull request to datman-config

DESm1th:
Hey Mike,

Just noticed the change to the TagInfo class, do all the scripts that call config.get_tags still work ok with the pattern and tag swapped? At a glance, dm_link_project_scans.py, dm_proc_fmry.py, dm_qc_report.py, and dm_to_bids.py might need a small update too

josephmje:
Thanks Dawn. I didn't realize those scripts used it as well. I've only updated `dm_link_sprl.py` so far.

josephmje:
I think the other scripts should be ok. Its only the ones that try to guess the tag from the series description that would be affected.

josephmje:
For the nifti conversion, if the filename already exists, the script skips transferring it. I noticed this was happening a lot of the dti scans. Apparently, some scanners produce an ADC file that's appended to the series. dcm2niix will discard them if you use the ignore feature (`-i y`). Should I add this flag to the script?

bin/dm_xnat_extract.py
```diff
@@ -687,50 +756,65 @@ def export_mnc_command(seriesdir, outputdir, stem):
...
-        return
+    logger.info("Exporting series {}".format(seriesdir))
 
-    logger.debug("Exporting series {} to {}".format(seriesdir, outputfile))
+    if multiecho:
+        echo_dict = get_echo_dict(stem)
 
     # convert into tempdir
     with datman.utils.make_temp_directory(prefix="dm_xnat_extract_") as tmpdir:
         datman.utils.run('dcm2niix -z y -b y -o {} {}'
```
josephmje: For the nifti conversion, if the filename already exists, the script skips transferring it. I noticed this was happening a lot for the dti scans. Apparently, some scanners produce an ADC file that's appended to the series. dcm2niix will discard them if you use the ignore feature (`-i y`). Should I add this flag to the script?

`dcm2niix -z y -b y -i y -o {} {}`

josephmje:
Also, dm2niix sometimes produces files with '_Eq' appended to them. For a 3D stack, it assumes that all the slices are equally spaced apart. But if they aren't the Eq file is made to try and interpolate the space in between. I haven't noticed any Eq files for the studies yet but it was a problem in POND. They were formed when we were missing some of the slices due to file transfer issues from the MR server. Should this situation be flagged in the script?

josephmje:
@edickie do you have any feedback about the '_ADC' and '_Eq' files?

bin/dm_xnat_extract.py
josephmje: Turns out the `-i y` is a bad idea. It sometimes skips scan tags that we would want like phase and phasediff images.

bin/dm_xnat_extract.py
```diff
@@ -378,41 +347,45 @@ def process_resources(xnat_project, session_label, experiment_label, data):
...
                          .format(xnat_resource_id, session_label, e))
             continue
 
         for resource in resources:
             resource_path = os.path.join(target_path, resource['URI'])
             if os.path.isfile(resource_path):
-                logger.debug('Resource:{} found for session:{}'
+                logger.debug("Resource: {} found for session: {}"
                              .format(resource['name'], session_label))
+                continue
```
jerdra: Is this necessary in if/else?

josephmje: good catch, I fixed this

bin/dm_link_sprl.py
```diff
@@ -137,12 +137,12 @@ def process_session(cfg, db, dir_nii, dir_res, session):
...
     # Doing it this way will enable matching multiple types of SPRL with
     # different regexs
-    sprls = [(i, v) for i, v in export_info.iteritems() if 'SPRL' in v]
+    sprls = [(i, v) for i, v in export_info.iteritems() if 'SPRL' in i]
 
     # find matching files in the resources folder
     sprl_files = []
     for sprl in sprls:
-        p = re.compile(sprl[0], re.IGNORECASE)
+        p = re.compile(sprl[1]['SeriesDescription'], re.IGNORECASE)
```
jerdra: A bit confused on this one, I thought series_map returns a dictionary mapping tags --> SeriesDescription,
so sprl should be a list of tuples (tag,seriesdescrip), how is accessing via key not raising an error?

bin/dm_xnat_extract.py
```diff
@@ -165,114 +171,114 @@ def main():
...
     xnat_project = session[0]
     session_label = session[1]
 
-    logger.info('Processing session:{}'.
-                format(session[1]))
+    logger.info("Processing session: {}".format(session_label))
+
+    # session_label should be a valid datman scanid
+    try:
+        ident = datman.scanid.parse(session_label)
```
jerdra: There's a bit of redundancy here with error handling  in collect_sessions for the ID validity and extra handling here. 
Unless they're checking different things, is there a way we could remove the redundant checks/handling?

bin/dm_link_sprl.py
josephmje: I modified the config.py so that it maps SeriesDescription --> tag. For some of the field maps and PDT2 scans, they would have the same series description but we'd want to map them to different tags.

bin/dm_xnat_extract.py
josephmje: Thanks. I didn't realize `datman.utils.validate_subject_id` was also doing a check. Perhaps we can pass the ident to the tuple made by `collect_sessions`.

bin/dm_link_sprl.py
jerdra: The changes were from:
` series_map[pattern] = tag`
to
` series_map[tag] = pattern`

That maps a tag --> series-description doesn't it? 
But either way I don't know where you're getting the 'SeriesDescription' key from which is what I'm confused about. Like I tried to replicate what you did by first updating my config.py to match yours then:
```
import datman.config as config
cfg = config.config(study='OPT')
tags = cfg.get_tags(site='WU1')
exportinfo = tags.series_map
```

So exportinfo is a dictionary mapping tags --> seriesdescriptions, I'm not seeing a 'SeriesDescription' key, am I missing something?

bin/dm_xnat_extract.py
josephmje: Nvm I think I know why the script had this. `collect_sessions` isn't run if we only specify a single session to extract.

jerdra: oh fair point, although it'd be nice to restructure it so collect_sessions doesn't check and let's process_session handle it, could do it later on! :+1:

bin/dm_xnat_extract.py
```diff
@@ -478,138 +444,296 @@ def process_scans(xnat_project, session_label, experiment_label, scans):
...
+        file_stem, tag, multiecho = create_scan_name(exportinfo,
+                                                     scan_info,
+                                                     session_label)
+        if not file_stem:
             continue
 
-        # first check if the scan has already been processed
+        if multiecho:
+            for stem, t in zip(file_stem, tag):
+                if (wanted_tags and t not in wanted_tags):
```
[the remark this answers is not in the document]
josephmje: Reading this back now was a little confusing for me as well haha. Fixed

bin/dm_link_sprl.py
josephmje: I changed the yaml as well. Here's an example:
```

Comment (jerdra) [reply bin/dm_link_sprl.py]: ah okay makes sense!

File: bin/dm_xnat_extract.py [hunk 6]
```diff
@@ -1,14 +1,15 @@
...
 Usage:
-    dm_xnat_extract.py [options] <study>
-    dm_xnat_extract.py [options] <study> <session>
+    dm_xnat_extract.py [options] <study> [--tag=<t>]...
+    dm_xnat_extract.py [options] <study> <session> [--tag=<t>]...
 
 Arguments:
     <study>            Nickname of the study to process
     <session>          Fullname of the session to process
+    --tag=<t>          List of scan tags to download
```
DESm1th: Is tag an option or an argument? It uses option syntax but looks like docopt will force at least one to be listed

josephmje: Tag is an option. I wasn't quite sure what the right syntax to use for this. It seems to work if I enter `dm_xnat_extract.py --tag=T1 --tag=T2` but that's pretty clunky.

jerdra: maybe have them enter as comma-separated list?

jerdra: Btw if you wanted tag as an option, then the syntax looks like:

`dm_bids_app.py [options] [-e <EXCLUDE>]... [-s <SUBJECT>]...  <study> <out> <json>`

Where -e and -s are repeated arguments

josephmje: Thanks! I'll modify it to that ^

DESm1th: Ah ok! Yeah it looks like that's about the best docopt can do for repeating options. Could you move the tag option to the 'Options:' section to make it clearer it's not a mandatory argument?

josephmje: Done

PR #175, opened 2018-11-02
Update dm_bids_app to handle longitudinal better with bids naming convention + dm_to_bids run multiple people without submitting

jerdra:


jerdra:
it looks like dm_to_bids changed a lot but actually i only changed the if else condition for the for-loop. Everything inside is the same!

bin/dm_bids_app.py
```diff
@@ -94,8 +94,13 @@ def get_bids_name(subject):
     try:
         ident = scan_ident.parse(subject)
     except scan_ident.ParseException:
-        logger.error('Cannot parse {} invalid DATMAN name!'.format(subject))
-        raise
+        logger.error('Cannot parse {}, attempting to reconcile with _01!'.format(subject))
+
+        try:
+            ident = scan_ident.parse(subject + '_01')
```
gabiherman: Does it make sense to add a warning or notification here that it reconciled w/ _01? I know it's slightly redundant, but might be helpful. Up to you.

jerdra: Good catch, some leftover stuff from debugging oops!

jerdra: done :)

PR #181, opened 2018-12-06
Add slurm to dm_bids_app

jerdra:
we'll need to export the SBATCH_PARTITION environment variable when using slurm btw


jerdra:
Added a few changes so that we could take advantage of pre-existing bids directories

PR #185, opened 2018-12-18
fix redcap and sprl scan symlinking

josephmje:


josephmje:
not sure why auto assign hasn't been working :S

DESm1th:
Yeah I tried fixing it yesterday (by fixing I mean uninstalling and re-installing, and then trying in vain to run it locally and test it before giving up :p) and I can't figure it out either. I'll happily do 1 review. Who do you feel like for your second? XD

PR #191, opened 2019-01-22
Patch 1

DESm1th:
Fixed all the noticeable hiccups from the previous changes. Login errors were because erin (and possibly others?) were using an old url to access the dashboard which github didnt like.

### DIFF
--- a/bin/dm_link_shared_ids.py
+++ b/bin/dm_link_shared_ids.py
@@ -213,8 +213,8 @@ def make_links(record):
             target_tags = target_cfg.get_tags(site=record.id.site).keys()
         except:
             target_tags = []
-        else:
-            target_tags = ",".join(target_tags)
+
+        target_tags = ",".join(target_tags)
 
         link_scans.create_linked_session(str(source), str(target), target_tags)
 
--- a/bin/dm_link_shared_ids.py
+++ b/bin/dm_link_shared_ids.py
@@ -125,7 +125,7 @@ def get_project_redcap_records(config, redcap_cred):
         if record.matches_study(current_study):
             project_records.append(record)
 
-    #Return list of records for selected studies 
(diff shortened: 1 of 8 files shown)


datman/dashboard.py
```diff
@@ -137,8 +137,15 @@ def add_subject(name):
 @dashboard_required
 @scanid_required
 def get_session(name, create=False, date=None):
+    try:
+        sess_num = datman.scanid.get_session_num(name)
+    except datman.scanid.ParseException:
+        logger.info("{} is missing a session number. Using default session "
+                "'1'".format(name))
```
gabiherman: Would this over-write an existing session 1 scan?

DESm1th: Good catch! it doesnt affect any of our scripts because we always call get_session with create=True but I'll make sure to fix this in the dashboard next round of bug fixes

bin/dm_link_shared_ids.py
```diff
@@ -208,8 +208,15 @@ def make_links(record):
     for target in record.shared_ids:
         logger.info("Making links from source {} to target {}".format(source,
                 target))
+        target_cfg = datman.config.config(study=target)
+        try:
+            target_tags = target_cfg.get_tags(site=record.id.site).keys()
+        except:
+            target_tags = []
+        else:
+            target_tags = ",".join(target_tags)
```
edickie: Why is there an else here? I'm confused.

DESm1th: If the tags are found they need to be a comma separated list. If an exception happens nothing needs to be done with it

DESm1th:
Ha, yeah. I think I was on too much of a 'try/except/else' kick since both you and gabi commented on it. It wasnt necessary there, I think.

PR #193, opened 2019-02-01
Added sagittal and coronal views for fmri and dti on qc pages

DESm1th:


DESm1th:
We can. It's the same command line dependency the multiple parts of the qc page already relied on so it's not a new issue, though. Should I look into this after the JSON header checks?

edickie:
Here's an example of me splitting a square image into one longer horizontal one with pillow (note height and width were specified here but I think there is a fast way to read from from the image with pillow too.
```
from PIL import Image
            with Image.open(tmp_img) as img:
                half_the_height = height // 2
                img_top = img.crop((0, 0, width, half_the_height))
                img_btm = img.crop((0, half_the_height, width, height))
                im2 = Image.new('RGBA', (int(width*2), half_the_height))
                im2.paste(img_top, (0, 0))
                im2.paste(img_btm, (width, 0))
                im2.save(output_loc)
```

PR #195, opened 2019-02-14
fix and simplify dm_bids_app with log-dir argument

jerdra:


assets/dm_bids_app_config/mriqc_local.json
```diff
@@ -0,0 +1,14 @@
+{
+
+	"app":"MRIQC", 
+	"img": "/archive/code/containers/MRIQC/poldracklab_mriqc_0.11.0-2018-06-05-1e4ac9792325.img",
+
+	"bidsargs":
```
gabiherman: Do we want to specify the fd_thresh at 0.5 since that is what we use in most of our studies? Or maybe have an option to specify that from command line if it isn't already straightforward to? (Sorry if it already is)

assets/dm_bids_app_config/mriqc_scc.json
```diff
@@ -0,0 +1,13 @@
+{
+
+	"app":"MRIQC", 
+	"img": "/KIMEL/tigrlab/archive/code/containers/MRIQC/poldracklab_mriqc_0.11.0-2018-06-05-1e4ac9792325.img",
+
+
+	"bidsargs": 
+	{
+		"n_procs":2
```
gabiherman: The local one has "verbose":"True"

Should that also be the case here?

assets/dm_bids_app_config/mriqc_local.json
jerdra: welll this might be problematic, it depends on the study..

gabiherman: Please update the mriqc container to the newest version (1.14 i think?)

assets/dm_bids_app_config/mriqc_scc.json
jerdra: sure i can update that, but not sure what you mean by local?

assets/dm_bids_app_config/mriqc_local.json
jerdra: going to think about how to add custom arguments via command-line, good catch

edickie: A fancier thing would be to read these from the project_settings instead of an overall config (although these could be the defaults?) Mathu is working on a project adding web form that might help with that.

I think I would +1 and the fd_thresh for 0.5 for all onging kimel data collection.

jerdra:
updated mriqc 0.11 --> 0.14

assets/dm_bids_app_config/mriqc_local.json
jerdra: @edickie I'll set it at 0.5 for now, then we can add an over-ride if needed. Although I'm not too happy about MRIQC throwing out the fd timeseries tbh

jerdra: > I think I would +1 and the fd_thresh for 0.5 for all onging kimel data collection.

1. Not sure what you're referring to the first bit, sorry can you clarify that? 

2. I'll add 0.5 as default for now, although I think it's something we should discuss - i think it might be better to get an idea of motion in each study before setting an arbitrary criteria (MRIQC should give us FD traces tbh)

3. Yes to project_settings, alternatively we could have different config files dependent on study. It's just tricky because you'd have to have sections in the project settings dependent on the specific bids-app. I don't know if throwing everything there is a sustainable solution.

assets/dm_bids_app_config/mriqc_scc.json
jerdra: oh nvm  i see lel

jerdra:
going to add an additional commit to handle slurm partitions as an option sorrrrry, this hotfix is needed

assets/dm_bids_app_config/mriqc_scc.json
jerdra: fixed!

jerdra:
ok I swear i'm done lol. We can discuss study-specific fd and other changes afterwards?

PR #198, opened 2019-02-21
Correct missing default_log variable

josephmje:
The only real change is on lines 649 and 650. The rest is automatic formatting from atom.


josephmje:
Sorry @DESm1th . I added a command to turn off uploading the mriqc metrics to their web server. That part was erroring out and the `filter_subjects` will rerun subjects if 'error' is in the log file.

PR #200, opened 2019-02-28
Fix dm_bids_app issue of MRIQC resubmitting jobs when already complete

jerdra:


bin/dm_bids_app.py
```diff
@@ -697,6 +684,78 @@ def group_subjects(subjects):
...
+        for key in keys:
+            if key in contents:
+                return True
+
+def error_in_mriqc(out_dir, log_file):
+    '''
+    Search target file for error keywords
+    '''
+
+    keywords = ['RuntimeError','ERROR','FileNotFoundError',
```
gabiherman: Have you thought about possible other ways to do this? In my experience mriqc generates a lot of log errors for stuff we don't really care about. (Maybe that isn't a change for right now, though.)

jerdra: just to keep a record of what i mentioned in slack:

MRIQC generates errors for connecting to their server, i specifically used these keywords to avoid the ones related to those types of warnings. 

Unfortunately we don't really have a good way of checking for true errors currently, maybe when bids derivatives come out there will be a systematic way to check - but for now we're kinda stuck with this workaround.

PR #201, opened 2019-02-28
Json standards

DESm1th:
Before pulling into the archive we need to:
   - Add  timepoints.header_diffs to the production database
   - Add a list of header fields to ignore to tigrlab_config.yaml under the key 'IgnoreHeaderFields'
   - Add a dictionary of header fields and their tolerances to tigrlab_config.yaml under 'HeaderFieldTolerance'.  

Still working on the dashboard UI changes to actually use this info, but the QC pages will work as before (I cleaned up the table that presents diffs a bit)

### DIFF
--- a/bin/dm_redcap_scan_completed.py
+++ b/bin/dm_redcap_scan_completed.py
@@ -67,7 +67,7 @@ def get_version(api_url, token):
 
 def add_session_redcap(record):
     record_id = record['record_id']
-    subject_id = record[cfg.get_key(['REDCAP_SUBJ'])].upper()
+    subject_id = record[cfg.get_key('REDCAP_SUBJ')].upper()
     if not datman.scanid.is_scanid(subject_id):
         try:
             subject_id = subject_id + '_01'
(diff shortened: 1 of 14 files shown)


bin/dm_redcap_scan_completed.py
```diff
@@ -157,8 +157,8 @@ def main():
     project_records = []
     for item in response.json():
         # only grab records where instrument has been marked complete
-        if not (item[cfg.get_key(['REDCAP_DATE'])] and
-                item[cfg.get_key(['REDCAP_STATUS'])] == '1'):
+        if not (item[cfg.get_key('REDCAP_DATE')] and
+                item[cfg.get_key('REDCAP_STATUS')] == '1'):
```
gabiherman: Just confirming - my PR will over-write this to use the new config variable, right?

DESm1th: Good catch! Turns out it causes a conflict. I resolved the conflict with the most recent commit

PR #202, opened 2019-03-01
study by study control of redcap scan complete value

gabiherman:
redcap_scan_completed now  checks the specified variable REDCAP_STATUS for value REDCAP_STATUS_VALUE

(please merge config before this)

### DIFF
--- a/bin/dm_redcap_scan_completed.py
+++ b/bin/dm_redcap_scan_completed.py
@@ -67,7 +67,7 @@ def get_version(api_url, token):
 
 def add_session_redcap(record):
     record_id = record['record_id']
-    subject_id = record[cfg.get_key(['REDCAP_SUBJ'])].upper()
+    subject_id = record[cfg.get_key('REDCAP_SUBJ')].upper()
     if not datman.scanid.is_scanid(subject_id):
         try:
             subject_id = subject_id + '_01'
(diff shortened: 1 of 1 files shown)


bin/dm_redcap_scan_completed.py
```diff
@@ -141,27 +141,35 @@ def main():
...
+
+        #make status_val into a list
+        if not (isinstance(status_val,list)):
+            status_val=[status_val]
+
         # only grab records where instrument has been marked complete
-        if not (item[cfg.get_key(['REDCAP_DATE'])] and
-                item[cfg.get_key(['REDCAP_STATUS'])] == '1'):
+        if not (item[cfg.get_key('REDCAP_DATE')] and
+                item[cfg.get_key('REDCAP_STATUS')] in status_val):
```
josephmje: Just to double check, are there any cases where a T1 wasn't collected but the other scans were? Would just checking for the presence of `REDCAP_STATUS` or the `mr_status` variable for OPT be enough?

```suggestion
                item[cfg.get_key('REDCAP_STATUS')]):
```

gabiherman: Jerry and I were discussing the best way to do this - we think that a T1 being collected is a more reliable indicator of scans being collected because the mr_status value hasn't been filled out by RAs in a consistent way.

PR #205, opened 2019-03-11
fix dm_qc_report bug on failing when getting to "missing" field

jerdra:


jerdra:
waiiiiit will resubmit soon

PR #208, opened 2019-03-18
Modified to work with new CMH FMAP tags

Mathu-M:


Mathu-M:
Yup! There's also the issue with the other fieldmap types (MAG1 and 2, PHASE1 and 2, and PHASEDIFF). Currently only topup and our ECHO1 and 2 fieldmaps are being handled using the dictionary in the get_intended_fors function since we know what files they should be associated with. We need to know that same information for the other fieldmap types and it would be best if if wasn't just held in a local dictionary. Both of these issues are going to be discussed tomorrow.

DESm1th:
Thanks for being on top of that! You're awesome :D

PR #213, opened 2019-04-29
Added exception for failed API request

gabiherman:
just a quick addition :)

### DIFF
--- a/bin/dm_redcap_scan_completed.py
+++ b/bin/dm_redcap_scan_completed.py
@@ -161,6 +161,11 @@ def main():
 
     response = get_records(api_url, token, instrument)
 
+    #http status code 200 indicates a successful request, everything else is an error.
+    if response.status_code != 200:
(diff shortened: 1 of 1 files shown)


bin/dm_redcap_scan_completed.py
```diff
@@ -161,6 +161,11 @@ def main():
 
     response = get_records(api_url, token, instrument)
 
+    #http status code 200 indicates a successful request, everything else is an error.
```
jerdra: What do we think about encapsulating the exception handling in get_records - the less main() knows about get_records the better is my perspective. Lemme know thoughts

DESm1th: Main is already breaking encapsulation a bit further down where it gets response.json(). So I would agree that it is better inside get_records but if so we should change get_records to return json itself too.

gabiherman: Let's do that, good suggestion.

PR #215, opened 2019-05-17
[WIP] make VRMD SPRL scans have different series numbers

josephmje:
open to any suggestions on how not to hardcode this.


DESm1th:
Hrm. I don't know if there's any way to avoid hard coding it somewhere. If it's not too much work though it may be more extensible/easier to modify if the hard coding happens in the settings yaml instead of the code itself. Like maybe having a special config block that, if present, causes it to match a series description in the settings to a number also specified in the settings and if not present do the 'else' block. That way we can just modify the settings if we discover this problem with any other studies going forward

josephmje:
Thanks @DESm1th . I'll think about how to port it over to the settings yaml.

PR #223, opened 2019-07-08
Add task file management script

DESm1th:
dm_task_files.py finds the task files in the resources folders using a regex (that may be overridden in the config files with the 'task_regex' key) and links them to their own folder. It also adds them to the dashboard database if the dashboard is configured.

### DIFF
--- a/bin/dm_task_files.py
+++ b/bin/dm_task_files.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+"""
+Usage:
+    dm_task_files.py <study>
+
+Arguments:
+    <study>             A datman managed study name
+"""
+
+import os
+import re
+import glob
+import logging
+
+from docopt import docopt
+
+import datman.config
+import datman.utils
+from datman.scan import Scan
+import datman.dashboard as dashboard
+
+logging.basicConfig(level=logging.WARN,
+        format="[%(name)s] %(levelname)s: %(message)s")
+logger = logging.getLogger(os.path.basename(__file__))
+
+def main():
+    args = docopt(__doc__)
+    study = args['<study>']
+
+    config = datman.config.config(study=study)
+    subjects = datman.utils.get_subject_metadata(config)
+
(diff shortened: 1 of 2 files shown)


bin/dm_task_files.py
```diff
@@ -0,0 +1,163 @@
...
+                add_to_dashboard(session, dest_path)
+
+
+def get_regex(config):
+    try:
+        regex = config.get_key('task_regex')
+    except datman.config.UndefinedSetting:
+        logger.warn("'task_regex' not defined in settings, using default regex "
+                "to locate task files.")
+        regex = 'behav|\.edat2'
```
gabiherman: Unfortunately the edat2 files don't always actually end with .edat2 because of RA error. (Regardless we want the txt file versions of the edat2 files). I wonder if (very unfortunately) we might need to call a command line utility to get the list of files for OPT. 

also should this be hard-coded?

DESm1th: If it's in a folder with any variation of 'behav' in the name it will still find it. It takes the entire contents of the 'behav' folders as task data, and just filters out any files that match the 'ignore' string. And it can be overridden by setting 'task_regex' in the study or site config files. That's just the default.

gabiherman: Ohhh oops. Cool! 

The task files are not always in behav - sometimes they are in a different folder within RESOURCES. But this happens rarely enough that it seems do-able for staff to move the files if they're in the wrong spot.

DESm1th: Yeah.. that's unfortunate. If they're elsewhere with some sort of consistent pattern we could just add to the regex 'behav|\.edat2|otherfolder' etc. If it's not... we'd have to do it manually no matter what, unfortunately :(

DESm1th:
Don't merge yet, I'm adding tsv generation now that Gabi reminded me about it yesterday

DESm1th:
Ok task tsv generation will come later because network debugging is my life now so it's ok to merge / review this for now.

PR #234, opened 2019-10-07
Add dm_parse_ea.py

gabiherman:
The very messy version of dm_parse_ea that was used to parse EA task log files in early October 2019. This will be eventually updated to not be hardcoded and to actually use datman. but for now just committing this to datman from my https://github.com/gabiherman/parse_ea_task so it can be found centrally. It would probably be better suited for rigorous code review a little later down the line...

### DIFF
--- a/bin/dm_parse_ea.py
+++ b/bin/dm_parse_ea.py
@@ -1,0 +1,294 @@
+#!/usr/bin/env python
+
+"""
+Parses SPINS' EA log files into BIDS tsvs
+usage:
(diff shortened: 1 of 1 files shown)


gabiherman:
As discussed with @DESm1th , I'd like to commit this to datman for the purpose of record-keeping. for now you can approve it, i'll make a PR with a better version later that is more suitable for code review.

josephmje:
Is this ready for review?

PR #235, opened 2019-10-10
Moar patches! (and some refactoring)

DESm1th:
This pull request:

1. Stops dm_symlink_scans from complaining when it looks for dicoms and they're missing because they were blacklisted.
2. Fixes the bug I accidentally introduced where dm_qc_report.py would crash when a json was missing or the gold standard wasnt found. The code for generating header diffs with the dashboard now behaves the same as the code for doing it from the file system.
3. Refactors dm_xnat_extract.py a bit to improve readability
4. Fixes a bug in dm_xnat_extract.py where some scans (definitely multi-echo scans but possibly others) werent having jsons added when they got exported because they didnt have a scan entry in the database in time for the call to 'update_side_cars'

### DIFF
--- a/bin/dm_qc_report.py
+++ b/bin/dm_qc_report.py
@@ -804,6 +804,7 @@ def run_header_qc(subject, config):
             return
         for series in db_session.scans:
             if not series.active_gold_standard:
+                header_diffs[scan_name] = {'error': 'Gold standard not found'}
                 continue
             check_bvals = needs_bval_check(tag_settings, series)
             if not series.json_contents:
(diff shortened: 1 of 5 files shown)


bin/dm_qc_report.py
```diff
@@ -804,14 +804,19 @@ def run_header_qc(subject, config):
             return
         for series in db_session.scans:
             if not series.active_gold_standard:
+                header_diffs[scan_name] = {'error': 'Gold standard not found'}
```
josephmje: `scan_name` is only defined in line 825

josephmje:
Just one comment to double check. Otherwise, LGTM!

bin/dm_qc_report.py
DESm1th: Ah thanks for catching that!

PR #241, opened 2019-11-01
Added a bids file class to scanid

DESm1th:
Added a basic bids file class to datman.scanid so I can more easily work with bids file names when finding / deleting files.

### DIFF
--- a/datman/scanid.py
+++ b/datman/scanid.py
@@ -35,10 +35,10 @@ BIDS_SCAN_RE = r'sub-(?P<subject>[A-Z0-9]+)_' + \
                r'(?P<suffix>(?!run-)[^_.]+)' + \
                r'.*$'
 
-SCANID_PATTERN = re.compile('^'+SCANID_RE+'$')
-SCANID_PHA_PATTERN = re.compile('^'+SCANID_PHA_RE+'$')
-FILENAME_PATTERN = re.compile('^'+FILENAME_RE)
-FILENAME_PHA_PATTERN = re.compile('^'+FILENAME_PHA_RE)
+SCANID_PATTERN = re.compile('^' + SCANID_RE+'$')
+SCANID_PHA_PATTERN = re.compile('^' + SCANID_PHA_RE+'$')
+FILENAME_PATTERN = re.compile('^' + FILENAME_RE)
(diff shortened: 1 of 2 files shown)


datman/scanid.py
```diff
@@ -19,35 +19,47 @@
...
-FILENAME_PATTERN     = re.compile('^'+FILENAME_RE)
-FILENAME_PHA_PATTERN = re.compile('^'+FILENAME_PHA_RE)
-
-#python 2 - 3 compatibility hack
+              r'(?P<ext>.nii.gz|.nii|.json|.bvec|.bval|.tar.gz|.tar|.dcm|' + \
+              r'.IMA|.mnc|.nrrd|$)'
+
+BIDS_SCAN_RE = r'sub-(?P<subject>[A-Z0-9]+)_' + \
+               r'ses-(?P<session>[0-9][0-9])_' + \
+               r'(run-(?P<run>[0-9])_){0,1}' + \
```
josephmje: The `task-` or `acq-` and `dir-` entities come before `run-`. Did you want to capture these optional entities as well?

DESm1th: Whoops! I didnt notice that was an issue because it wasnt causing anything to throw a parse exception when I tested it on all the bids file names in the database. Let me fix this (thanks for noticing Mike!)

josephmje: I think the priority list is:
task-
acq-
ce-
dir-
rec-
run-
echo-

DESm1th: Thanks Mike! I added them in and added some more tests

DESm1th:
Erin reminded me that bids doesnt require the session number to be a number. So I made one last update (sorry!).

datman/scanid.py
```diff
@@ -30,7 +30,7 @@
               r'.IMA|.mnc|.nrrd|$)'
 
 BIDS_SCAN_RE = r'sub-(?P<subject>[A-Z0-9]+)_' + \
-               r'ses-(?P<session>[0-9][0-9])_' + \
+               r'ses-(?P<session>[A-Za-z0-9]+)_' + \
```
josephmje: Thanks for catching this!

PR #245, opened 2019-12-17
[REF] Update code to Python 3 + PEP8 changes + remove deprecated scripts

DESm1th:

PR #247, opened 2019-12-18
[FIX POST #245] - dm_task_files to pull all subjects instead of just checklisted ones.

jerdra:
Small tiny bug to fix so that task files are actively scraped even for unchecked files

## Requires

- [ ] https://github.com/TIGRLab/dashboard/pull/51
- [x] https://github.com/TIGRLab/datman/pull/245


jerdra:
update: integrating additional changes to use dashboard mechanism since get_subjects will be deprecated in #245

jerdra:
We should merge this PR after #245.

jerdra:
Also an additional dependency on https://github.com/TIGRLab/dashboard/pull/51 being integrated

PR #248, opened 2019-12-30
[FIX] Modify scanid variable in ``dm_link.py``

josephmje:
Was getting the following error message from `dm_link.py`:

```
  File "/archive/code/datman/datman/utils.py", line 1051, in validate_subject_id
    except scanid.ParseException:
UnboundLocalError: local variable 'scanid' referenced before assignment
```


jerdra:
thanks for catching this!

PR #252, opened 2019-12-31
[FIX] Python3 conversion bug fixes (Round 2)

DESm1th:


bin/dm_qc_report.py
```diff
@@ -504,8 +504,8 @@ def get_series_to_add(series, subject):
         t2 = get_split_image(subject, series.series_num, 'T2')
```
josephmje: I think we can do away with this bit. `dm_xnat_extract.py` now takes care of splitting the PDT2 scan into separate files.

jerdra:
looks good to me!

bin/dm_qc_report.py
jerdra: might be out of scope of PR, could maybe do a separate one for this specific change?

josephmje: yup i'm fine with that.

DESm1th: Ooh good catch. I'll roll that into another one soon (if someone doesnt beat me to it)

PR #253, opened 2020-01-02
[STY] PEP8 style changes and minor logging fixes

josephmje:


datman/xnat.py
```diff
@@ -161,10 +161,10 @@ def get_project(self, project):
         return(result['items'][0])
 
     def get_sessions(self, study):
-        logger.debug('Querying xnat server for sessions in study'
+        logger.debug('Querying xnat server for sessions in study:{}'
```
DESm1th: If it's not a giant hassle (or a difference in preferred style) could you throw a space after the colons in these messages? I dont know about you but all our old log messages that leave the space out drive me a bit crazy haha

DESm1th:
Thanks for catching all these issues!

datman/xnat.py
josephmje: Will do!

jerdra:
looks good to me, thanks for catching!

jerdra:
will approve after pep8 changes!

josephmje:
> will approve after pep8 changes!

i was going to ask whether we should increase the line length?

DESm1th:
Hmm.. I personally dont have any strong preferences either way. But I will say that python's actual line length limit lets me fit two tabs side by side on that lab monitors without missing the end of lines or having to close atom's project pane haha. I can definitely work around that though and will just go with whatever you guys prefer

jerdra:
Agree with @DESm1th on editor usability. 
Messing around with line lengths will always be weird to deal with :man_shrugging:  since you kind of want to always add that extra bracket over the end. 
Maybe we'll just exercise judgement on whether we think a particular flake8 issue is problematic? 

For example, for the current pep8 issues I'd be totally cool with just letting it pass since it's just a bracket.

DESm1th:
^ All of that sounds incredibly reasonable to me (including letting the bracket pass).

jerdra:
Okkkk Dawn and I had a mini-chat:
We propose the following:

Use 10% feature on the condition that all auto-code formatters should stick to 80 characters. Dawn and I have been using yapf for our code-formatting which we're mostly happy with.

That way any violation of 80 characters is intentional and made with full awareness

josephmje:
Cool! I'm adding some extra configuration to flake8 in my WIP PR. Will keep line length at 80 and add B950.

PR #258, opened 2020-01-14
[REF] Clean up error messages

DESm1th:
No one seems to care about the multiple UUIDs so I've made that quieter, and I've allowed dm_blacklist_rm to delete blacklisted scans even if the entire session hasn't finished being signed off.

### DIFF
--- a/bin/dm_link_shared_ids.py
+++ b/bin/dm_link_shared_ids.py
@@ -171,9 +171,9 @@ def link_shared_ids(config, connection, record):
     experiment = get_experiment(subject)
 
     if not experiment:
-        logger.error("Redcap record may be misnamed - no matching "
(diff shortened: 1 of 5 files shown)


bin/dm_link_shared_ids.py
```diff
@@ -171,7 +171,8 @@ def link_shared_ids(config, connection, record):
     experiment = get_experiment(subject)
 
     if not experiment:
-        logger.error("No matching experiments for subject {}. "
+        logger.error("Redcap record may be misnamed - no matching "
```
gabiherman: Do you think this might be more informative if it says REDCap or XNAT might be misnamed, since it could be either?

DESm1th: Good point! Let me fix that

PR #260, opened 2020-02-03
[ENH/MAINT] Implement CircleCI

jerdra:
# Major Changes

Built up CircleCI config borrowing heavily from Niprep's build/test/deploy model with minor modifications for Datman.

Particularly important is that *any* commits directly to master (releases qualify as this) will cause a deploy. From looking at Niprep's implementation pull requests don't trigger deployment since they are evaluated as non-master branches. 

**Basically don't commit to Master directly.**

Lmk if you have any feedback/suggestions!

## Additional Notes

1. Using CodeCov for dealing with coverage reports

2. Was originally using tox to automate testing of multiple python versions but that complicated dealing with coverage across multiple python versions and reproducing distribution vs testing env so went for the approach of using the 3.6.5 as our base python to ensure 3.6+ compliance. 

3. Added a smol update to use dateutils parser to mitigate issues with TZ parsing

## Additional to-dos

- [x] Code coverage is being handled by https://circleci.com/orbs/registry/orb/codecov/codecov which handily deals with code coverage uploading. We will have to enable third-party repos in our CircleCI
- [x] Update CircleCI ENV VAR for CodeCov token


.circleci/config.yml
```diff
@@ -1,41 +1,176 @@
...
+          name: Check if docs build before continuing
+          command: |
+            cd ~/datman
+            if [[ "$( git log --format='format:%s' -n 1 $CIRCLE_SHA1 | grep -iE '^docs?(\(\w+\))?:' )" != "" ]]; then
+              echo "Will only be building Docs"
+              circleci step halt
+            fi
       - restore_cache:
-          key: deps1-{{ .Branch }}-{{ checksum "/home/circleci/datman/requirements.txt" }}
+          key: datman-{{ .Branch }}-{{ checksum "setup.cfg" }}
```
josephmje: can we add a version to this (eg. datman-v1-)? just in case we need to update the cache down the line.

.circleci/config.yml
```diff
@@ -1,41 +1,176 @@
...
       - run:
-          name: Install requirements in venv
+          name: Set up installation and cache
           command: |
-            python -m virtualenv venv
-            source venv/bin/activate
-            pip install -r /home/circleci/datman/requirements.txt
+            # Install datman under python 3.6.1 with test packages
+            pyenv global 3.6.5
```
gabiherman: Comment says 3.6.1, you load 3.6.5? Which version do you want?

.circleci/config.yml
```diff
@@ -1,41 +1,176 @@
...
+            echo "CIRCLETAG: $CIRCLEC_TAG"
+            echo "VERSION: \"$THISVERSION\""
+            echo "INSTALLED: \"$INSTALLED_VER\""
+            test "$INSTALLED_VER" = "$THISVERSION"
+      - run:
+          name: Check wheel
+          command: |
+            python -m venv /tmp/whl
+            source /tmp/whl/bin/activate
+            pip install pyyaml wrapt
```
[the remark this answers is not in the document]
jerdra: they already are, the issue is that during this stage of building the source distribution we haven't actually installed any packages from setup.cfg yet. When get_version gets called in the next run block it calls __init__ which fails if wrapt/pyyaml aren't loaded in (presumably because datman proceeds to load additional modules from __init__). 

Lmk if you think otherwise!

.circleci/config.yml
jerdra: the cache will be updated anytime setup.cfg is modified for each branch independently. What other invalidation conditions are you thinking of for site-packages?

.circleci/config.yml
jerdra: vestigal comment from before i discovered circleci classic doesn't have 3.6.1. I'll remove it

.circleci/config.yml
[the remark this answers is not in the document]
josephmje: ahhh right because the config is getting loaded by `__init__.py`

.circleci/config.yml
josephmje: i've just been seeing this in other circleci configurations and assumed it was some sort of best practice. we could also add multiple keys as a failsafe

```
keys:
    - datman-{{ .Branch }}-{{ checksum "setup.cfg" }}
    - datman-{{ .Branch }}-
    - datman-
```

.circleci/config.yml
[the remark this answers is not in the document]
jerdra: what I can do is to grep the pyyaml/wrapt version directly from setup.cfg, that could be a decent compromise

.circleci/config.yml
jerdra: Did a bit of searching around and I concede on both your points. 

Would be good to have a mechanism to manually invalidate/rebase the cache and using vX achieves that. The fall back is a good idea as well, i don't think there are any conditions in which we specifically require a clean install w/respect to site-packages.

PR #262, opened 2020-02-07
[FIX/MAINT] - Only deploy on pypi when commits/PRs are made to the release branch

jerdra:
Since **Master** will be our in-house production branch which is always being tweaked, we want to reserve deployment for milestones/major bug fixes.

Any commit/PR pushed to the branch **release** will be deployed to pypi.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -179,7 +179,6 @@ workflows:
             - deployable
           filters:
             branches:
-              only: release
+              ignore: /.*/
             tags:
               only: /.*/
-
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -170,7 +170,7 @@ workflows:
             - test_deploy_pypi
           filters:
             branches:
-              only: release
+              only: master
             tags:
               only: /.*/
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -170,7 +170,7 @@ workflows:
             - test_deploy_pypi
           filters:
             branches:
-              only: master
+              only: release
```
josephmje:
```suggestion
              only: master
```

.circleci/config.yml
```diff
@@ -179,7 +179,7 @@ workflows:
             - deployable
           filters:
             branches:
-              only: master
+              only: release
```
josephmje:
```suggestion
              ignore: /.*/
```

jerdra:
> I think we want `test_deploy_pypi` to run at every PR because if our package doesn't build, there could be something very wrong.

This is already the case. 

> Also, we don't currently have a release branch. I think `deployable` is fine to run on the master branch but only for tagged releases. 

Why? Deployable triggers subsequent workflows for deployment - the idea is to make a release branch that will always have code ready to go. 

> For some reason, that tags regex isn't getting picked up?

I think it's because /.*/ picks up on empty tags as well (CIRCLECI_TAG) is empty. Maybe something like <code>\d+\.\d+\.\d+(-rc\d+)?</code> could be a suitable regex? That way we could keep it on master and it would only release on the aforementioned tag regex.

PR #265, opened 2020-03-06
[ENH] Add KCNI ID support

DESm1th:
This updates our core scripts that interact with XNAT (and dm_link) to allow zip files and xnat experiments to use the KCNI conventions. 

Note: I did **NOT** yet update dm_link_shared_ids.py, dm_xnat_project_overview.py, or track_scan_dates.py. The first needs a major overhaul (it still doesnt allow SENDEP linking to work right, and predates the dashboard) that I'd like to spend more time on and none of the three will be affected yet by our initial KCNI XNAT test cases. But I'll push that later since we need to move forward.

For studies and sites that we want to start migrating we'll have to update the settings file to add:

XNAT_CONVENTION: KCNI

and for study / site fields that need to be changed for KCNI we'll have to add an id map that matches this format:

ID_MAP:
  STUDY: {KCNI_STUDY: DATMAN_STUDY}
  SITE: {
      KCNI_SITE1: DM_SITE1, 
      ...
      KCNI_SITEN: DM_SITEN
  }

### DIFF
--- a/bin/dm_link.py
+++ b/bin/dm_link.py
@@ -10,18 +10,18 @@ Usage:
     dm_link.py [options] <study> <zipfile>
 
 Arguments:
-    <study>              Name of the study to process
-    <zipfile>           Single Zipfile to process
+    <study>                 Name of the study to process
+    <zipfile>               Single Zipfile to process
 
 Options:
-    --lookup FILE            Path to scan id lookup table,
-                                overrides metadata/scans.csv
-    --scanid-field STR       Dicom field to match target_name with
-                             [default: PatientName]
-    -v --verbose             Verbose logging
-    -d --debug                  Debug logging
-    -q --quiet             Less debuggering
-    --dry-run             Dry run
+    --lookup FILE           Path to scan id lookup table,
+                            overrides metadata/scans.csv
+    --scanid-field STR      Dicom field to match target_name with
+                            [default: PatientName]
+    -v --verbose            Verbose logging
+    -d --debug              Debug logging
+    -q --quiet              Less debuggering
+    --dry-run               Dry run
 
 
 DETAILS
--- a/bin/dm_link.py
+++ b/bin/dm_link.py
@@ -70,17 +70,21 @@ IGNORING EXAM ARCHIVES
         source_name      target_name            dicom_StudyID
         2014_0126_FB001  <ignore>
 """
-from docopt import docopt
-import sys
-import os
+
 import glob
-import pandas as pd
-import datman.config
-import datman.utils
-import datman.scanid
 import logging
+import os
+import sys
+
+from docopt import docopt
+import pandas as pd
+
+import datman.config
+import datman.scanid
+import datman.utils
 
 logger = logging.getLogger(os.path.basename(__file__))
+
 already_linked = {}
 lookup = None
 DRYRUN = None
@@ -93,14 +97,14 @@ def main():
     global DRYRUN
 
     arguments = docopt(__doc__)
-    verbose = arguments['--verbose']
-    debug = arguments['--debug']
-    DRYRUN = arguments['--dry-run']
-    quiet = arguments['--quiet']
-    study = arguments['<study>']
-    lookup_path = arguments['--lookup']
-    scanid_field = arguments['--scanid-field']
-    zipfile = arguments['<zipfile>']
+    verbose = arguments["--verbose"]
+    debug = arguments["--debug"]
+    DRYRUN = arguments["--dry-run"]
+    quiet = arguments["--quiet"]
+    study = arguments["<study>"]
+    lookup_path = arguments["--lookup"]
+    scanid_field = arguments["--scanid-field"]
+    zipfile = arguments["<zipfile>"]
 
     # setup logging
     ch = logging.StreamHandler(sys.stdout)
@@ -116,8 +120,8 @@ def main():
         logger.setLevel(logging.DEBUG)
         ch.setLevel(logging.DEBUG)
 
-    formatter = logging.Formatter('%(asctime)s - %(name)s - {study} - '
-                                  '%(levelname)s - %(message)s'.format(
+    formatter = logging.Formatter("%(asctime)s - %(name)s - {study} - "
+                                  "%(levelname)s - %(message)s".format(
                                                     study=study))
     ch.setFormatter(formatter)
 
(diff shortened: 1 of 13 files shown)


DESm1th:
I felt like I wrote so many new tests. And then I codecov says +2.31% coverage. *Cries internally*

datman/scanid.py
```diff
@@ -92,13 +43,147 @@ def get_full_subjectid_with_timepoint_session(self):
...
+
+    .. note: The original KCNI format ID is retrievable from orig_id.
+
+    """  # noqa: E501
+
+    scan_re = '(?P<id>(?P<study>[A-Z]{3}[0-9]{2})_' \
+              '(?P<site>[A-Z]{3})_' \
+              '(?P<subject>[A-Z0-9]{4,8})_' \
+              '(?P<timepoint>[0-9]{2})_' \
+              'SE(?P<session>[0-9]{2})_MR)'
```
JayHennessy: This looks good, although Im not sure how datman handles experiments with incorrect naming. Im just thinking that in the future, they will not all have modality code "MR". If experiments that fail this regex match are ignored, then this is good.

bin/dm_link.py
```diff
@@ -193,33 +197,36 @@ def link_archive(archive_path, dicom_path, scanid_field, config):
...
-        datman.utils.validate_subject_id(scanid, config)
-    except RuntimeError as e:
-        logger.error(str(e) + ". Cannot make link for {}".format(archive_path))
+        ident = datman.utils.validate_subject_id(scanid, config)
+    except datman.scanid.ParseException as e:
+        logger.error("Can't make link for {}. Reason: {}".format(
+            archive_path, e))
         return
 
+    scanid = str(ident)
```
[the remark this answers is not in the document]
DESm1th: scanid is overwritten to ensure that scanid is a datman style ID. For instance, if scanid was a valid KCNI ID using the str(ident) would set scanid to be the datman version of that ID. For datman IDs there's no difference between scanid and str(ident) though. 

Ident is assigned if validation passes, otherwise there's a ParseException and the function logs and exits :) So it's safe to reassign on line 218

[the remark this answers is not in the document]
JayHennessy: Ok, sounds good!

datman/scanid.py
DESm1th: It should raise a ParseException whenever there's not a match, and currently all our scripts just log and exit. So it should be ok (possibly spammy though, haha)

bin/dm_xnat_upload.py
```diff
@@ -122,76 +113,115 @@ def is_datman_id(archive):
...
+        try:
+            settings = CFG.get_key("ID_MAP")
+        except datman.config.UndefinedSetting:
+            settings = None
+        try:
+            scanid = datman.scanid.get_kcni_identifier(scanid, settings)
+        except datman.scanid.ParseException:
+            logger.error("ID {} can't be converted to KCNI convention.".format(
+                scanid))
+            return
```
jerdra: Noticing that getting the scanid always requires this kind of dance, would it be a good idea to move this chunk into scanid and let it be handled there?

DESm1th: Possibly. getting / checking the xnat convention should probably stay out since it's only relevant in scripts that interact with xnat. Getting the settings could be, but as it is now it makes it a tiny bit easier to implement unit tests for get_kcni_identifier and makes the tests a bit more readable.

datman/xnat.py
```diff
@@ -88,260 +89,399 @@ class xnat(object):
...
+                the search to. Defaults to None.
+
+        Returns:
+            str: The name of the XNAT project the subject belongs to. Note:
+                if the same ID is found in more than one project only the
+                first match is returned.
+        """
+        if not projects:
+            projects = self.get_projects()
+            projects = [p["ID"] for p in projects]
```
jerdra: What about:

```suggestion
            projects = [p["ID"] for p in self.get_projects()]
```

datman/xnat.py
```diff
@@ -88,260 +89,399 @@ class xnat(object):
...
         except Exception:
-            raise XnatException("Failed getting xnat sessions with url: {}"
+            raise XnatException("Failed getting xnat subjects with URL {}"
                                 .format(url))
 
         if not result:
-            raise XnatException('No sessions found for study: {}'.format(study))
+            return []
+
+        return [item.get("label") for item in result["ResultSet"]["Result"]]
```
jerdra:
```suggestion
        return [item["label"] for item in result["ResultSet"]["Result"]]
```
To keep consistent with lookup usage w/result, afaik .get is used moreso for the second parameter (unless item is not a dict, in which case ignore me :speak_no_evil: )

datman/xnat.py
```diff
@@ -88,260 +89,399 @@ class xnat(object):
...
-        if result is None:
-            e = XnatException('Scan not found for experiment: {}'
-                              .format(experiment))
-            e.study = study
-            e.session = session
-            raise e
+        if not result:
+            return []
+
+        return [item.get("ID") for item in result["ResultSet"]["Result"]]
```
jerdra:
```suggestion
        return [item["ID"] for item in result["ResultSet"]["Result"]]
```
same as earlier suggestion

datman/xnat.py
```diff
@@ -825,101 +950,98 @@ def _make_xnat_delete(self, url, retries=3):
...
-class Session(object):
+    def __repr__(self):
+        return self.__str__()
+
+
+class XNATObject(ABC):
+    def _get_field(self, key):
+        if not self.raw_json.get("data_fields"):
+            return ""
+        return self.raw_json["data_fields"].get(key, "")
```
jerdra: could not get_field just be a function taking a XNAT object json + key? Doesn't look like the ABC is being used for much atm

datman/xnat.py
```diff
@@ -351,24 +491,24 @@ def get_resource_ids(self, study, session, experiment, folderName=None,
...
+        for r in result["ResultSet"]["Result"]:
             try:
-                label = r['label']
-                resource_ids[label] = r['xnat_abstractresource_id']
+                label = r["label"]
+                resource_ids[label] = r["xnat_abstractresource_id"]
             except KeyError:
                 # some resource folders have no label
-                resource_ids['No Label'] = r['xnat_abstractresource_id']
+                resource_ids["No Label"] = r["xnat_abstractresource_id"]
```
jerdra: What about:

```
label = r.get("label", "No Label")
resource_ids[label] = r["xnat_abstractresource_id"]
```

datman/xnat.py
DESm1th: Ah the reason I did it that way is that XNAT (appears) to always return a dictionary with the keys ResultSet and Result but the content of "Result" doesn't necessarily always have the label defined. I'm not 100% certain about that though (XNAT doesnt really document return json structure and isnt always consistent). I'll change it and catch/reraise KeyErrors as XNATExceptions for consistency

datman/xnat.py
DESm1th: Urf, missed that one when I was cleaning up old functions. Nice catch!

datman/xnat.py
DESm1th: True, I wrote that, I think, when I started reworking the classes and then there was less code in common to pull out than I expected. I think the benefit of having it as an abstract base class though is that we can always easily refactor things into XNATObject if we do find more commonalities later

datman/xnat.py
DESm1th: Whoops! Yeah that's way better, I didnt actually read it when I fixed the quotes hahaha

bin/dm_xnat_extract.py
```diff
@@ -178,85 +169,93 @@ def configure_logging(study, quiet=None, verbose=None, debug=None):
...
                 continue
-
-            if (not datman.scanid.is_phantom(session['label']) and
-                    sub_id.session is None):
-                logger.error("Invalid ID {} in project {}. Reason: Not a "
-                             "phantom, but missing series number"
-                             .format(session['label'], project))
+            if (ident.session is None and not datman.scanid.is_phantom(ident)):
+                logger.error("Invalid experiment ID {} in project {}. Reason "
+                             "- Not a phantom, but missing series number"
```
josephmje: Should this be series number or visit/session?

bin/dm_xnat_upload.py
```diff
@@ -99,15 +90,15 @@ def main():
...
         else:
-            logger.error('Cant find archive:{}'.format(archive))
+            logger.error("Cant find archive:{}".format(archive))
             return
     else:
         archives = os.listdir(dicom_dir)
 
-    logger.debug('Processing files in:{}'.format(dicom_dir))
-    logger.info('Processing {} files'.format(len(archives)))
+    logger.debug("Processing files in:{}".format(dicom_dir))
```
josephmje:
```suggestion
    logger.debug("Processing files in: {}".format(dicom_dir))
```

datman/scanid.py
josephmje: We could theoretically add multiple possible modality codes once we have them.

bin/dm_xnat_extract.py
DESm1th: Whoops, nice catch!

bin/dm_xnat_upload.py
DESm1th: Update for the record - We talked and agreed it's not great but there's not really another clean solution (without a different set of drawbacks) that we could come up with so we'll both think on it some more and update it later

PR #266, opened 2020-03-10
DOCS: Self-host docs using GitHub pages

josephmje:
## Changes proposed in this PR
- add Sphinx configuration
- update CircleCI config to automatically update docs
- string formatting to Python 3 f-strings
- automatic code formatting as a results of the following VSCode Settings

```
{
  "python.pythonPath": "/Users/michaeljoseph/.pyenv/versions/datman_venv/bin/python",
  "python.linting.flake8Enabled": true,
  "python.linting.flake8Args": [
    "--config=setup.cfg"
  ],
  "python.formatting.provider": "black",
  "python.formatting.blackArgs": [
    "--line-length=80"
  ],
  "python.sortImports.args": [
    "-sp setup.cfg"
  ],
  "[python]": {
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    }
  },
  // Configure editor settings to be overridden for [yaml] language
  "[yaml]": {
    "editor.insertSpaces": true,
    "editor.tabSize": 2,
  }
}
```

## To Do
- [ ] check that fingerprints are correct in CircleCI config


josephmje:
docs are being built here! https://96-29540972-gh.circle-artifacts.com/0/docs/_build/no_version_html/html/index.html

datman/utils.py
```diff
@@ -341,49 +388,65 @@ def read_blacklist(study=None, scan=None, subject=None, config=None, path=None,
...
+            config=config,
+            bids_ses=bids_ses,
+            use_bids=use_bids,
+        )
 
     if use_bids:
-        raise MetadataException("Can't return BIDs blacklist info without "
-                                "dashboard integration")
+        raise MetadataException(
+            "Can't return BIDs blacklist info without " "dashboard integration"
```
DESm1th, line 402: Not a big deal, but this should probably be one string if on one line now

docs/requirements.txt
```diff
@@ -1,13 +1,13 @@
...
-pybids==0.10.0
-pydicom==1.3.0
-pysftp==0.2.9
-pyxnat==1.2.1.0.post3
-pyyaml==5.2
-requests==2.22.0
-urllib3==1.25.7
-wrapt==1.11.2
+git+https://github.com/AleksandarPetrov/napoleon.git@0dc3f28a309ad602be5f44a9049785a1026451b3#egg=sphinxcontrib-napoleon
+git+https://github.com/rwblair/sphinxcontrib-versioning.git@39b40b0b84bf872fc398feff05344051bbce0f63#egg=sphinxcontrib-versioning
```
DESm1th, line 2: Is there a reason to use these specific commits on these repos instead of just the plain old sphinxcontrib-napoleon and sphinxcontrib-versioning?

datman/utils.py
josephmje: thanks for finding this. will fix!

docs/requirements.txt
josephmje: the napoleon one has a feature for rending the Parameters header in the docstring that hasn't been merged yet. but I'll remove the versioning one.

josephmje:
> Nice work! Thanks for putting this together and pushing us toward using the newer style f strings. I think moving forward we should definitely use them, but maybe make an exception and use .format for really large strings though. For me at least it's easy to lose track of variables needed to fill out a large string and accidentally delete or rename something, but format helps a bit in these cases by keeping them grouped at the end. What do you think? Maybe it doesn't make much a difference (it shouldn't if we have tests over code in these instances, haha).

Thanks Dawn. I made the switch the improve readability and cut down on some of the line lengths. But I used an automatic converter to do it and wasn't discriminating between really large strings. That's a good point. I'm in favour of any option that would help with understanding the code.

DESm1th:
Yeah! I don't think we actually have very many strings large enough for it to matter and none of the strings you updated are all that long. More tests is definitely a better solution than deciding on a case by case basis whether to switch to the better f-string format haha.

PR #267, opened 2020-03-12
[FIX] Bug fixes from adding KCNI name support

DESm1th:
This fixes a problem where scan tags with the substring "PHA" were being flagged as invalid, and adds tests to catch this and other similar tag issues, and also reverts some of the changes made to xnat_fetch_sessions because I forgot that their server's experiments are not meaningful and it isn't really necessary to get them to switch to KCNI format.

### DIFF
--- a/datman/scanid.py
+++ b/datman/scanid.py
@@ -114,7 +114,7 @@ class DatmanIdentifier(Identifier):
         return self.get_full_subjectid_with_timepoint_session()
 
     def get_xnat_experiment_id(self):
-        return self.xnat_subject_id()
+        return self.get_xnat_subject_id()
 
     def __repr__(self):
         return '<datman.scanid.DatmanIdentifier {}>'.format(self.__str__())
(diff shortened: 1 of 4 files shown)


datman/scanid.py
```diff
@@ -68,7 +68,7 @@ class DatmanIdentifier(Identifier):
 
     scan_re = '(?P<id>(?P<study>[^_]+)_' \
               '(?P<site>[^_]+)_' \
-              '(?!.+PHA)(?P<subject>[^_]+)_' \
+              '(?P<subject>[^_]+)(?<!PHA)_' \
               '(?P<timepoint>[^_]+)_' \
               '(?!MR)(?!SE)(?P<session>[^_]+))'
```
josephmje: is the MR and SE here a hack for dealing with phantoms?

DESm1th: That and to ensure KCNI IDs (correctly formed or not) won't match. We can't just restrict to numeric sessions only because of prelapse (and possibly others).

PR #272, opened 2020-04-08
[FIX] Experiment renaming failures

DESm1th:
So apparently some experiments on XNAT can end up with more than one autorun pipeline instance (even though the history in the GUI never shows more than one) and the other instance(s) can hang around and make things fail. I thiiiink this might only happen with older data from before we upgraded XNAT (can't confirm but so far only older experiments have had the problem).


jerdra:
how annoying :see_no_evil:

PR #273, opened 2020-04-28
[FIX] Small but fatal xnat.py errors

DESm1th:
This fixes:

1. A reference to (the undefined) self.experiment that's currently preventing download of zip files
2. A type error that causes crashes when multiple sessions are present on XNAT


DESm1th:
Er.. hold off on this one for a bit, I think there may still be bugs to squash.

DESm1th:
Forgot to update, @jerdra @edickie  but this appears to be ok to review as is (no other errors cropped up while downloading prelapse).

bin/xnat_fetch_sessions.py
```diff
@@ -126,12 +126,12 @@ def download_subjects(xnat, xnat_project, destination):
...
-            return
+            continue
 
         if len(exp_names) > 1:
             logger.error("Found {} experiments for subject {}".format(
                 len(exp_names), subject.name))
-            return
+            continue
 
         experiment = subject.experiments[exp_names[0]]
```
jerdra, line 136: Might be a good idea to indicate that additional experiment names are being ignored in a log message (first one is being selected only)? Unless I'm misunderstanding the intention here :speak_no_evil:

bin/xnat_fetch_sessions.py
```diff
@@ -147,7 +147,7 @@ def download_subjects(xnat, xnat_project, destination):
             logger.info("Would have downloaded experiment {} from project "
                         "{} to {}".format(
                             experiment.name, xnat_project, zip_path))
-            return
+            continue
```
jerdra, line 150: shouldn't the dry-run not proceed w/making tmp dirs and downloading data?

jerdra:
> Forgot to update, @jerdra @edickie but this appears to be ok to review as is (no other errors cropped up while downloading prelapse).

hey sorry this totally fell through the cracks!

bin/xnat_fetch_sessions.py
jerdra: so far all the subject MR sessions have been separate subject IDs. I was just pointing out here that if they had done that, then the behaviour of the script wouldn't be obvious from a log

DESm1th: This was just because when there's more than one experiment on their server it's usually indicative of other naming issues too. I've updated the log message now so it indicates the entire subject is being skipped.

bin/xnat_fetch_sessions.py
DESm1th: It will skip that part, but I accidentally let it make the destination directory if it didnt exist so I've fixed that now, thanks Jer!.

DESm1th:
Sorry it took me so long to get back to this @jerdra , @josephmje . It kind of took a back seat to all the KCNI transfer stuff

josephmje:
Thanks @DESm1th ! There's still some conflicts

datman/xnat.py
```diff
@@ -1205,23 +1211,20 @@ def _make_xnat_post(self, url, data, retries=3, headers=None):
...
+                                    "dicom upload fail?")
             else:
-                raise XnatException(
-                    "An unknown error occurred uploading data."
-                    f"Status code: {response.status_code}, "
-                    f"reason: {response.content}"
-                )
-        return response.content
+                raise XnatException("An unknown error occured uploading data."
+                                    "Status code: {}, reason: {}"
```
josephmje, line 1224:
```suggestion
                                    f"Status code: {response.status_code}, "
```

datman/xnat.py
```diff
@@ -1205,23 +1211,20 @@ def _make_xnat_post(self, url, data, retries=3, headers=None):
...
+                                    .format(response.status_code,
```
josephmje, line 1225:
```suggestion
                    f"reason: {reply}")
```

datman/xnat.py
```diff
@@ -1205,23 +1211,20 @@ def _make_xnat_post(self, url, data, retries=3, headers=None):
...
+                                            reply))
```
josephmje, line 1226:
```suggestion

```

josephmje (approved):

jerdra (approved):

PR #281, opened 2020-07-22
Bidsify fix - don't re-write dataset_description.json on every run

JayHennessy:
In bin/bidsify.py I just added a check to see if dataset_description.json exists, then don't re-write it if it does.

I also wrote a quick unittest for this for this addition.


JayHennessy:
Oops, i didn't do a pep8 check. Just made the updates now.

PR #286, opened 2020-09-24
Added parser for Nback and GNGo

ThomasHMAC:


bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,128 @@
...
+import pandas as pd
+import subprocess
+import glob
+import sys
+
+arguments       = docopt(__doc__)
+inputfile        = arguments['<inputfile>']
+resultsfile      = arguments['--output']
+DEBUG           = arguments['--debug']
+DRYRUN          = arguments['--dry-run']
```
jerdra, line 40: can you run a linter/code formatter for this? Dawn and I have been using "yapf" to re-format our code!

bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,128 @@
...
+
+if DEBUG : print(arguments)
```
jerdra, line 42: can you indent here? running a re-formatter could work as well

bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,128 @@
...
+
+## loop over the inputfiles
+
+if DEBUG: print("\n file: {}".format(inputfile))
```
jerdra, line 46: same as above re:indenting

bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,128 @@
...
+import codecs
+import sys
+import pdb
+from docopt import docopt
+import numpy as np
+import os
+import pandas as pd
+import subprocess
+import glob
+import sys
```
jerdra, line 34: Might be a good idea to use "import logging" to print to stdout. Would be worth checking other datman/bin scripts and seeing how they log!

bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,128 @@
...
+        if (("ImageDisplay1.RT:" in line) and ("ImageDisplay1.RTTime:" not in line)):
+            ReactionTime = line.split(':')[1].replace(" ",'')
+        if "ImageDisplay1.RESP:" in line:
+            SubjectResponse = line.split(':')[1].replace(" ",'')
+        if "ImageDisplay1.CRESP:" in line:
+            CorrectResponse = line.split(':')[1].replace(" ",'')
+        if "image:" in line:
+            StimFile = line.split(':')[1].replace(" ",'')
+    #Skip if list type
+    if (TrialType == ''):continue
```
jerdra, line 87: indent here

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+## loop over individual trials
+trial_count = 0
+for trial in trials:
+
+    # for each trial collect the CorrectResponse, Subject's Response and the Reation Time
+    CorrectResponse = int()
+    ReactionTime = int()
+    SubjectResponse = int()
+    TrailType = str()
+    TrailTime = int()
```
jerdra, line 68:
```suggestion
    TrialType = str()
    TrialTime = int()
```

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+
+    for line in trial.split('\n\t'):
+        if ("Running:" in line) :
+            TrailType = line.split(':')[1].replace(" ",'')
```
jerdra, line 72:
```suggestion
            TrialType = line.split(':')[1].replace(" ",'')
```

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+        if ("letterdisp.OnsetTime:" in line) :
+            TrailTime = line.split(':')[1].replace(" ",'')
```
jerdra, line 74:
```suggestion
            TrialTime = line.split(':')[1].replace(" ",'')
```

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+        if ("resp:" in line) and ("letterdisp.RESP" not in line) and ("letterdisp.CRESP" not in line):
+            CorrectResponse = line.split(':')[1].replace(" ",'')
+        if (("letterdisp.RT" in line) and ("letterdisp.RTTime" not in line)):
+            ReactionTime = line.split(':')[1].replace(" ",'')
+        if "letterdisp.RESP" in line:
+            SubjectResponse = line.split(':')[1].replace(" ",'')
+
+    #Skip if list type
+    if (TrailType == 'List1') or (TrailType == ''):continue
```
jerdra, line 83:
```suggestion
    if (TrialType == 'List1') or (TrialType == ''):continue
```

Indent would be needed here as well

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+import glob
+import sys
+
+arguments       = docopt(__doc__)
+inputfile        = arguments['<inputfile>']
+resultsfile      = arguments['--output']
+DEBUG           = arguments['--debug']
+DRYRUN          = arguments['--dry-run']
+
+if DEBUG : print(arguments)
```
jerdra, line 37: indent here

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+            ReactionTime = line.split(':')[1].replace(" ",'')
+        if "letterdisp.RESP" in line:
+            SubjectResponse = line.split(':')[1].replace(" ",'')
+
+    #Skip if list type
+    if (TrailType == 'List1') or (TrailType == ''):continue
+    trial_count += 1
+
+    #Form trial dictionary
+    trial_dict = {'onset' : TrailTime, 'duration' : 0,
```
jerdra, line 87:
```suggestion
    trial_dict = {'onset' : TrialTime, 'duration' : 0,
```

bin/dm_parse_Nback.py
```diff
@@ -0,0 +1,119 @@
...
+                 'trial_type' : TrailType, 'correct_response' : corr[str(CorrectResponse)],
```
jerdra, line 88:
```suggestion
                 'trial_type' : TrialType, 'correct_response' : corr[str(CorrectResponse)],
```

jerdra (changes requested):
Thanks for setting this up Thomas! Just a few stylistic comments I've added. Would recommend that you try a linter/code formatter like yapf to adjust the code style to match what we've set out as code style standards.

Other thing to consider maybe is whether we can refactor these scripts into a single script/library of scripts to handle task parsing since there're a ton of overlaps b/w the two scripts. Obv having something working for the moment is important - but something to think about for sure!

ThomasHMAC:
> Thanks for setting this up Thomas! Just a few stylistic comments I've added. Would recommend that you try a linter/code formatter like yapf to adjust the code style to match what we've set out as code style standards.
> 
> Other thing to consider maybe is whether we can refactor these scripts into a single script/library of scripts to handle task parsing since there're a ton of overlaps b/w the two scripts. Obv having something working for the moment is important - but something to think about for sure!

Thanks for the suggestion Jer, and yes there are tons of overlapping between two scripts. I will try to make a single script/ library of scripts later

bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,132 @@
...
+    print(arguments)
+
+# Loop over the inputfiles
+if DEBUG:
+    print("\n file: {}".format(inputfile))
+
+try:
+    with open(inputfile, "r", encoding="utf-16") as myfile:
+        filetext = myfile.read()
+except myfile.DoesNotExist:
```
DESm1th, line 44: This should be updated to 

```
except FileNotFoundError:
```

DESm1th (changes requested):
Looks really good, thank you Thomas! Like Jer I just have a few stylistic python-nitpicks, but I'm comfortable approving things as they are now once the exception is fixed. 

1. In python it's good to use snake-case for multi-word variable and function names so for instance "inputfile" would be "input_file" instead. This is a PEP8 practice, but most linters can't check for it.

2. It's a good idea to put code inside a main function / other functions (and then call the main function at the bottom). Every instruction that's not inside of a function gets executed when the script is imported, which can make it difficult or impossible to write tests for code. If it's in a function, though, it doesn't execute until the function is called (with a standard 'guard' at the bottom of the script like below, the main function will always be called if the script is being run directly on the command line and not if it's imported somewhere else)
```
def main():
    # your code here

if __name__ == "__main__":
    main()
```

bin/dm_parse_GNGo.py
```diff
@@ -0,0 +1,132 @@
...
+message_df["response_time"] = message_df["response_time"].apply(
+    lambda x: x / 1000
+)
+
+# Load in the gngo onset time csv file
+try:
+    df_onset = pd.read_csv("gng_event_timing.csv")
+except df_onset.DoesNotExist:
+    print("ERROR opening onset time file")
+    quit()
```
jerdra, line 123: exception should be fixed here as well FileNotFoundError

jerdra (changes requested):

DESm1th (approved):
Thanks Thomas! Good work :D

jerdra (approved):

PR #287, opened 2020-09-25
[ENH] Allow the subid field to be modified between naming conventions

DESm1th:
To change the subject ID field of an ID add something formatted like the below example to the ID_MAP in a study's settings

```yml
ID_MAP:
    SUBJECT:
        'kcni_regex->replacement_str': 'dm_regex->replacement_str'
```

### DIFF
--- a/bin/dm_xnat_extract.py
+++ b/bin/dm_xnat_extract.py
@@ -227,6 +227,8 @@ def collect_all_experiments(config):
                                  "Reason - Not a phantom, but missing session "
                                  "number".format(exper_id, project))
                     continue
+                if ident.modality != "MR":
+                    continue
                 experiments.append((xnat, project, ident))
 
     return experiments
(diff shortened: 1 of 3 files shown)


josephmje (approved):
This looks good to me @DESm1th! I can see the kcni_regex possibly being annoying to code but this will do. Have you come across any other studies other than PACTMD where we would need to remap the subject id? If so, we might want to include more tests for those patterns.

DESm1th:
NEUR also needs it, not sure yet about other studies

jerdra:
ooo yeah these seem a bit difficult to deal with :vomiting_face: special cases :vomiting_face:... is there an alternative way to store this as SUBJECT --> KCNI = <regex, replace> and SUBJECT --> DATMAN = <regex, replace> as list elements instead of encoding stuff directly into the <key> and <value>? 

This is defs nitpicking tbf

DESm1th:
I thought about doing something like that but in a lot of places it would mean extra code to determine whether to grab the KCNI or datman entry, possibly in the calling script itself.

jerdra:
always one step ahead of me :laughing:

jerdra (approved):

PR #288, opened 2020-09-28
[FIX] Removed XNAT integration (was unused)

DESm1th:
We havent used the XNAT integration for linked IDs since the dashboard was made, and now that we're moving studies to the KCNI server it is causing crashes so I've removed it.

### DIFF
--- a/bin/dm_link_shared_ids.py
+++ b/bin/dm_link_shared_ids.py
@@ -103,7 +103,7 @@ def get_redcap_records(config, redcap_cred):
     try:
         project_records = parse_records(response, current_study, id_map)
     except ValueError as e:
-        logger.error("Couldnt parse redcap records for server response {}. "
+        logger.error("Couldn't parse redcap records for server response {}. "
                      "Reason: {}".format(response.content, e))
         project_records = []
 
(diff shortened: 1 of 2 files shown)


bin/dm_link_shared_ids.py
```diff
@@ -123,7 +96,12 @@ def get_project_redcap_records(config, redcap_cred):
...
     try:
-        project_records = parse_records(response, current_study)
+        id_map = config.get_key('ID_MAP')
+    except UndefinedSetting:
+        id_map = None
+
+    try:
+        project_records = parse_records(response, current_study, id_map)
     except ValueError as e:
         logger.error("Couldnt parse redcap records for server response {}. "
```
josephmje, line 106:
```suggestion
        logger.error("Couldn't parse redcap records for server response {}. "
```

josephmje (approved):
Thanks @DESm1th! Looking good to me. Only question I had is whether there's a reason for switching the logging from StreamHandler to basic config? I don't have any preference on this. I've just been following what was done in the past.

DESm1th:
It's just less verbose. basicConfig just adds a streamhandler.

jerdra:
looks much cleaner now! thanks for the refactor Dawn :1st_place_medal:

jerdra (approved):

josephmje (approved):

PR #290, opened 2020-11-02
[ENH] - Implement flexible JSON-based exclusion criteria for bidsify

jerdra:
This PR lets bidsify.py to implement more flexible decision tree-like behaviour using scan JSON metadata and the ExportSettings section of config. For example:

```
ExportSettings:
  T1:
    formats: ['nii','dcm','mnc']
    qc_type: anat
    bids:
      class: anat
      modality_label: T1w
      exclude:
        - and:
          - Manufacturer: Siemens
          - not:
              ImageType: NORM
```

Will allow for exclusion of Siemens scans that don't have "NORM" under ImageType.


josephmje:
Thanks for this @jerdra! I think we might already have a way to deal with this in `dm_xnat_extract.py` for properly classifying the fieldmaps. Have you tried adjusting the ExportInfo to be something like:
```
            T1:       { Pattern: { SeriesDescription: 't1_mprage_eyes_closed$',
                                          ImageType: '*NORM*' },
                          Count: 1 }
```

Unless the preference is to still download both T1 and T2 files but only BIDS convert one of them?

jerdra:
whaaaat? i had no idea! @DESm1th can you confirm?

DESm1th:
Looks like you already found it, but the ImageType config setting is used by datman/xnat.py in the _set_fmap_tag function.

jerdra:
@josephmje it won't work out of the box, but i think some adjustments to config can make it work...

I.e we'll need to create a second type for "T1-NORM", then spec a BIDS config for T1-NORM to add acq-prescannorm. Then using a BIDS filter file we should be able to fun fmriprep selectively while maintaining all outputs in /bids

PR #291, opened 2020-11-10
[ENH] - Remove single fmap-type assumption for BIDSIFY, more flexible derivative specification

jerdra:
## Changes:

- Can now have multiple fmap types determined by the (modality, acq) tuple. These will be independently matched to scans
- "inherit" spec under "alt" implemented allowing source files to augment the bids fields of their derived children
- Fixed case where if the derivative of a file is under a different class, then the destination directory needs to be updated
- When derivatives were specified they were selected in lieu of the parent scan. The update now allows you to specify derivatives in addition to the parent by simply specifying the derivative to have a type that is the parent's type. This means you can have the parent in addition to its derivative pushed into the BIDS directory instead of just the latter. (Useful for boldref/sbref cases)
- `is_ref` option added which denotes that the scan is a reference scan and that BIDS info about this particular scan needs to be supplemented with info from the next scan (task)

Examples:

```yaml
SBREF:{...

bids:{
    is_ref: true,
    alt:[
         {
             template: "<TEMPLATE>",
             type: FMRI-DPA,
             inherit: {acq: sbref},
         },
         {
              type: SBREF
         }
        ]
}
```

This specifies:
- SBREF as a reference scan (task info is supplemented by the next scan)
- SBREF has a derivative of type FMRI-DPA, in addition the child will inherit the acq-sbref property
- SBREF will also have itself as a child so that in addition to the child being pushed, it itself will also be pushed

## FIX

- Grouped scans no longer assigned to separate fmaps due to the series number alone being used for mapping to fmaps. 

This means, for example, a series of resting state scans will always be matched to the same fmap and not to different ones if it happens-so that the last resting state scan series number was closer to another fmap

### DIFF
--- a/bin/bidsify.py
+++ b/bin/bidsify.py
@@ -557,9 +557,7 @@ def make_dataset_description(bids_dir, study_name, version):
             json.dump({
                 "Name": study_name,
                 "BIDSVersion": version
-            },
-                      f,
-                      indent=3)
+            }, f, indent=3)
 
     return
 
--- a/bin/bidsify.py
+++ b/bin/bidsify.py
@@ -42,7 +42,6 @@ from datman.bids.check_bids import BIDSEnforcer
 
 from collections import namedtuple
 from itertools import groupby, product
-from operator import attrgetter
 
 # Set up logger
 logging.basicConfig(level=logging.WARN,
@@ -594,9 +593,7 @@ def make_dataset_description(bids_dir, study_name, version):
             json.dump({
                 "Name": study_name,
                 "BIDSVersion": version
-            },
-                      f,
-                      indent=3)
+            }, f, indent=3)
 
     return
 
--- a/bin/bidsify.py
+++ b/bin/bidsify.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-
 """
 This copies and converts files in nii folder to a bids folder in BIDS format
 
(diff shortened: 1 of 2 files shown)


jerdra:
will fix PEP issues tmrw!

jerdra:
@DESm1th @josephmje @benselby ready to review!

jerdra:
I just dreamt up a bug w/how the series numbers are used to match fmaps. Gonna push a small adjustment to that algo and we should be good...

i really need to write tests for bidsify one day..

DESm1th:
Haha, thanks for being thorough Jer! <3

jerdra:
thanks for the suggestion @DESm1th, i've updated the matching algo and ran some test data through with success. It's ready to review!

@josephmje @benselby

bin/bidsify.py
```diff
@@ -170,25 +170,34 @@ def update_source(self, cfg, be):
         # Process each alternative
         alts = []
         for d in alt:
+            '''
+            Allow "self" to propogate itself if the tag itself
```
DESm1th, line 174: I think this comment got a bit mangled haha

bin/bidsify.py
```diff
@@ -385,13 +396,12 @@ def pair_with(x):
     return pair_list
 
 
-def calculate_average_series(series_list):
-
+def get_first_series(series_list):
     """
     For each iterable of BIDSFiles calculate the average series number
```
DESm1th, line 401: I think this comment isnt accurate anymore

DESm1th (approved):
Looks good Jer! Thanks for updating this :)

bin/bidsify.py
jerdra: hahaha wellppp thanks for catching should be fixed :)

bin/bidsify.py
jerdra: updated!

DESm1th (approved):

bin/bidsify.py
```diff
@@ -202,6 +211,10 @@ def update_source(self, cfg, be):
             derivsfile.source = new_source
             derivsfile.bids = new_bids
             derivsfile.json = derivsfile._load_json(new_json)
+            derivsfile.dest_dir = os.path.abspath(
+                os.path.join(derivsfile.dest_dir, os.pardir,
+                             derivsfile.spec['class']))
+
```
josephmje, line 217:
```suggestion
                os.path.join(derivsfile.dest_dir, os.pardir(derivsfile.spec['class'])))

```

josephmje (approved):

PR #295, opened 2021-01-28
[FIX] Prevent error log messages from being lost

DESm1th:
In the dashboard config (https://github.com/TIGRLab/dashboard/blob/master/config/logging.py) I set datman logging to critical to reduce log spam a while back. It turns out that this setting was quietly taking effect on all the scripts in `datman/datman` always, because config imports part of the dashboard. 

So! If we instead set it to warn in `datman/__init__.py`, we still get less logging when the dashboard is running or imported but we get our error messages back when datman components are imported (even if datman imports the dashboard)


jerdra:
Thanks for updating this Dawn!

jerdra (approved):

jskocic (approved):

PR #297, opened 2021-02-11
Add a function to remove button press during scanner response 

ThomasHMAC:


ThomasHMAC:
I found another small issue in the code I need to address before PR

PR #298, opened 2021-02-12
Updated EA parse to remove button press during scan response 

ThomasHMAC:


bin/dm_parse_ea.py
```diff
@@ -48,24 +49,74 @@ def read_in_logfile(path):
...
+    # Remove the button press registered during scanner responses
+    scan_response = ["101", "104"]
+    indexes_to_drop = []
+
+    for index, row in log_file.iterrows():
+        if ("rating" in log_file["Code"][index]) and any(
+            resp in log_file["Code"][index - 2] for resp in scan_response
+        ):
+            indexes_to_drop.append(index)
+            indexes_to_drop.append(index - 2)
```
jerdra, line 62: Can you add a comment describing why the current index and 2 indices beforehand need to be dropped? Mostly so we don't have a magic number in there that people in the future wouldn't understand!

bin/dm_parse_ea.py
```diff
@@ -48,24 +49,74 @@ def read_in_logfile(path):
...
+
+    if len(indexes_to_drop) == 0:
+        log_file_cleaned = log_file
+    else:
+        log_file_cleaned = log_file.drop(log_file.index[indexes_to_drop])
+        log_file_cleaned = log_file_cleaned.reset_index(drop=True)
+        logger.warning(
+            "Removed registered rating occurred before or after actual rating"
+        )
```
jerdra, line 71: Maybe include the number of "chunks" being dropped here for QC purposes (i.e if an abnormally large number of indices got dropped). 

```suggestion
        logger.warning(
            f"Removed {len(indexes_to_drop)/2} registered ratings that occurred before or after actual rating"
        )
```

bin/dm_parse_ea.py
```diff
@@ -48,24 +49,74 @@ def read_in_logfile(path):
...
+
+    indexes_to_drop_1 = []
+
+    for index, row in log_file_cleaned.iterrows():
+        if ("rating" in log_file_cleaned["Code"][index]) and any(
+            resp in log_file_cleaned["Code"][index - 1]
+            for resp in scan_response
+        ):
+            indexes_to_drop_1.append(index)
+            indexes_to_drop_1.append(index - 1)
```
jerdra, line 81: I think a "why" comment is needed here as well, might just be me, but I'm not really sure why we dropped i, i-2, and now i,i-1!

bin/dm_parse_ea.py
```diff
@@ -48,24 +49,74 @@ def read_in_logfile(path):
...
+
+    if len(indexes_to_drop_1) == 0:
+        final_log_file = log_file_cleaned
+    else:
+        final_log_file = log_file_cleaned.drop(
+            log_file_cleaned.index[indexes_to_drop_1]
+        )
+        final_log_file = final_log_file.reset_index(drop=True)
+        logger.warning("Removed rating registered followed scanner responses")
```
jerdra, line 90: Same as above logging comment re: including the number of ratings dropped

bin/dm_parse_ea.py
```diff
@@ -48,24 +49,74 @@ def read_in_logfile(path):
...
+        }
+    )
+
+    df["trial_type"] = df["movie_name"].apply(
+        lambda x: "circle_block" if "cvid" in x else "EA_block"
+    )
+    df["duration"] = df["movie_name"].apply(
+        lambda x: int(vid_info[x]["duration"]) * 10000
+        if x in vid_info
+        else "n/a"
```
jerdra, line 114: maybe it might be a good idea to use to use `pd.NA` here to avoid having a mixed data-type column

bin/dm_parse_ea.py
```diff
@@ -48,24 +49,74 @@ def read_in_logfile(path):
...
+    )
+
+    df["stim_file"] = df["movie_name"].apply(
+        lambda x: vid_info[x]["stim_file"] if x in vid_info else "n/a"
```
jerdra, line 118: same as above `pd.NA` comment

bin/dm_parse_ea.py
```diff
@@ -332,17 +410,18 @@ def parse_task(ident, log_file, dest_dir, length_file, timing_file):
     if outputs_exist(log_file, output_path):
         return
 
-    # Reads in the log, skipping the first three preamble lines
+    # Reads in and clean the log, skipping the first three preamble lines
     try:
         log = read_in_logfile(log_file)
+        log_cleaned = clean_logfile(log)
```
jerdra, line 416: could this be outside the try/catch block? It looks like `clean_logfile` is mostly if not entirely safe

bin/dm_task_files.py
```diff
@@ -18,65 +18,83 @@
...
-            try:
-                os.mkdir(dest_folder)
-            except OSError:
-                pass
+        session = os.path.basename(resource_folder)
+        dest_folder = os.path.join(out_dir, session)
+        try:
+            os.mkdir(dest_folder)
+        except OSError:
+            pass
```
jerdra, line 52:
```suggestion
        os.makedirs(dest_folder, exist_ok=True)
```

Unless you want it to error out if intermediate directories don't exist!

jerdra (changes requested):

jerdra (approved):

josephmje (approved):

PR #300, opened 2021-03-26
Create tay_mripicture.py

kyjimmy:
This script is designed for TAY cohort study and creates a cropped T1 image for custom t-shirt.

### DIFF
--- a/bin/tay_mripicture.py
+++ b/bin/tay_mripicture.py
@@ -1,0 +1,41 @@
+#!/usr/bin/env python
+
+"""
+Creates MRI axial pictures for custom t-shirt for TAY cohort study. 
+"""
+
+import glob
+import matplotlib.pyplot as plt
+import numpy as np
+
+# Specify Participant ID
+SUBJ = 'TAY01_CMH_P99999_01'
+
+# Input and Output Directories
+INDIR = '/external/rprshnas01/tigrlab/archive/data-2.0/TAY/qc/'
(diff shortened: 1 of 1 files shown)


bin/tay_mripicture.py
```diff
@@ -0,0 +1,41 @@
...
+
+import glob
+import matplotlib.pyplot as plt
+import numpy as np
+
+# Specify Participant ID
+SUBJ = 'TAY01_CMH_P99999_01'
+
+# Input and Output Directories
+INDIR = '/external/rprshnas01/tigrlab/archive/data-2.0/TAY/qc/'
```
DESm1th, line 15: To fix hard-coding your script could import datman.config and retrieve the study QC folder by doing something like:

```python
config = datman.config.config(study=study_name)
qc = config.get_path("qc")
```

DESm1th (changes requested):
Hey Jimmy! 

Thanks for putting this together. Before merging it in it might be a good idea to remove the hard-coding. The best way to do this would be to use docopt to take the subject ID + study name + output directory from the command line. Their documentation is here http://docopt.org/ You can also look at some of our other bin scripts to get a sense for how we use it.

Also, you might have noticed some things were flagged by pep8speaks. This bot catches formatting issues to try to keep our code somewhat consistent with python's PEP8 coding styles. There are also command line tools to help check your formatting yourself. If you run pycodestyle or flake8 they can help you catch things before you commit.

PR #302, opened 2021-04-26
Create dm_mripicture.py

kyjimmy:
Fixed the hard coding issue and changed appearance of the output picture.


bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def main():
+    arguments = docopt(__doc__)
+    study = arguments['<study>']
+    participant = arguments['<participant>']
+    output = arguments['<output>']
+
+    config = datman.config.config(study=study)
+    outdir = '/projects/jwong/tshirt'
```
josephmje, line 53: get rid of hard-coded path
```suggestion
```

bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/env python
+"""
+Creates MRI axial pictures for custom T-shirt.
+
+Usage:
+    mripicture.py [options] <study>
+    mripicture.py [options] <study> <participant>
+    mripicture.py [options] <study> <participant> <output>
```
josephmje, line 8: I think adding the ellipsis will allow `participant` to be a repeating argument. Please double check this.

```suggestion
    mripicture.py [options] <study> <participant>...
    mripicture.py [options] <study> <participant>... <output>
```

bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
...
+import glob
+import matplotlib.pyplot as plt
+import nibabel as nib
+import numpy as np
+from docopt import docopt
+
+import datman.config
+
+
+def image_cropping(imgpath):
```
josephmje, line 29: Have you considered nilearn's [`plot_anat` function](https://nilearn.github.io/modules/generated/nilearn.plotting.plot_anat.html#nilearn.plotting.plot_anat)? It will also allow you to specify certain slices or views. It also automatically adjusts the vmin and vmax depending on the input image.

niworkflows also has a [`cuts_from_bbox` function](https://github.com/nipreps/niworkflows/blob/5792f74b24f529590584b36be8df07a5833d963e/niworkflows/viz/utils.py#L143) to automatically select slices at equal distances apart and avoid getting slices outside of the brain.

bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
...
+
+    if not os.path.exists(outdir):
+        os.makedirs(outdir)
+
+    for subject in subs:
+
+        # Set Path
+        nii_dir = config.get_path('nii')
+        nii_dir = os.path.join(nii_dir, subject, '*Sag-MPRAGE-T1.nii*')
+        imgpath = glob.glob(nii_dir)[0]
```
josephmje, line 75: You might run into a situation where there are multiple T1s. Should this script always select the first T1 or loop through and just create images for all of them?

bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
...
+        outdir = output
+
+    if not os.path.exists(outdir):
+        os.makedirs(outdir)
+
+    for subject in subs:
+
+        # Set Path
+        nii_dir = config.get_path('nii')
+        nii_dir = os.path.join(nii_dir, subject, '*Sag-MPRAGE-T1.nii*')
```
josephmje, line 74: It's been a while since I worked with datman. @DESm1th, is there a better way to grab the T1s here using the config tags?

josephmje (commented):
Looking pretty good @kyjimmy! Thanks for this addition. I left a few minor suggestions and some things to think through.

bin/dm_mripicture.py
kyjimmy: If I want to have a default output directory, where should I save the output, e.g. archive/data/{project}/data?

bin/dm_mripicture.py
kyjimmy: I just need one picture with the best MR quality for tshirt-printing. I am thinking to use the latest file only.

```
        nii_dir = config.get_path('nii')
        nii_dir = os.path.join(nii_dir, subject, '*Sag-MPRAGE-T1.nii*')
        
        # Find All T1 Files
        imgpath = glob.glob(nii_dir)

        # Find the Latest T1 File
        imgpath = max(imgpath, key=os.path.getctime) 
```

bin/dm_mripicture.py
josephmje: Yes, you could make a tshirt folder within data

bin/dm_mripicture.py
kyjimmy: I am hoping to create a picture like this: [MRI_sample.pdf](https://github.com/TIGRLab/datman/files/6379877/MRI_sample.pdf). 

However, `plot_anat` doesn't give me the options to pull the middle 3 sagittal slices. I tried plotting.plot_anat({nifti path},display_mode='x',cut_coords=3,threshold='auto') but it showed me slices of the whole brain. Same for the  `cuts_from_bbox` function. 

Is it okay to keep what I have here?

josephmje: i think `plot_anat(<nifti_img>, display_mode="x", cut_coords=(80, 101, 10), annotate=False, draw_cross=False, threshold="auto") will get you there. 

i'm fine with keeping it as is. my suggestion for going with nilearn was to help with cases where the image contrast is different (`threshold="auto"`) and adapting `cuts_from_bbox` to decide on the slices that would go in `cut_coords`

josephmje: if you're looking for something even simpler, you could also take the middle saggital slice (find the shape of the data and divide by 2). and then add 10 slices up or down for the other 2 panels.

bin/dm_mripicture.py
jerdra: you could:

1. use `datman.scan` w/the subject ID to get a `Scan` instance
2. `get_tagged_nii` to retrieve the T1 `Series` object
3. Then sort by `Series.series_num` to pick out the last T1 scan of the session

bin/dm_mripicture.py
jerdra: i would maybe just let `participant` be a repeatable `[option]`. It's much easier to use unix pipes that way (i.e `xargs` pipes) and it simplifies the user interface. 

Such as:
```
mripicture.py [options] [-s <SUBJECT>]... <study>
```

Same goes for `output`, if it's optional, then specify it as an optional argument rather than positional

bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
...
+    img = np.transpose(t1_slices, (0, 2, 1))
+    img = np.moveaxis(img, 0, 1)
+    img = np.reshape(img, (256, -1))
+    return img
+
+
+def get_all_subjects(config):
+    qc_dir = config.get_path('nii')
+    subject_qc_dirs = glob.glob(os.path.join(qc_dir, '*'))
+    all_subs = [os.path.basename(path) for path in subject_qc_dirs]
```
jerdra, line 42: just nitpicking a bit here! I'd rename `qc_dir` to `nii_dir`

bin/dm_mripicture.py
```diff
@@ -0,0 +1,92 @@
...
+    else:
+        subs = get_all_subjects(config)
+        print('Creating brain pictures for all subjects for',
+              study, 'project.')
+
+    if output:
+        outdir = output
+
+    if not os.path.exists(outdir):
+        os.makedirs(outdir)
```
jerdra, line 68:
```suggestion
   os.makedirs(outdir, exist_ok=True)
```

Its effectively the same behaviour, but just a tad cleaner :)

bin/dm_mripicture.py
jerdra: If using this method I'd maybe add a `-t TAG` option since some studies have multiple T1 variants (i.e `T1-NORM`)

jerdra (commented):
Good stuff Jimmy! Just a few additional comments from me

bin/dm_mripicture.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python
+"""
+Creates MRI axial pictures for custom T-shirt.
+
+Usage:
+    mripicture.py [options]... <study>
```
josephmje, line 6:
```suggestion
    mripicture.py [options] <study>
    mripicture.py [options] <study> [-s <subject>]...
    mripicture.py [options] <study> [-s <subject>]... [-t <tag>]
```

bin/dm_mripicture.py
```diff
@@ -0,0 +1,78 @@
...
+        scan = datman.scan.Scan(subject, config)
+        tagged_scan = scan.get_tagged_nii(tag)
+        idx = np.argmax([ss.series_num for ss in tagged_scan])
+
+        # Set Path
+        imgpath = tagged_scan[idx].path
+        outpath = os.path.join(outdir, subject + '_T1.pdf')
+
+        # Output Image
+        plotting.plot_anat(imgpath, cut_coords=(-20, -10, 2), display_mode='x',
```
josephmje, line 68: You might have to save the result of this to a variable and run `savefig` on that variable. The way you currently have it might give you a blank image. Like:

```
t1_pic = plotting.plot_anat()
t1_pic.savefig()
```

bin/dm_mripicture.py
```diff
@@ -0,0 +1,78 @@
...
+    config = datman.config.config(study=study)
+
+    if subs:
+        print('Creating brain pictures for subjects [', ', '.join(subs),
+              '] from', study, 'project.')
+    else:
+        subs = get_all_subjects(config)
+        print('Creating brain pictures for all subjects for',
+              study, 'project.')
+
```
josephmje, line 51: Like you do with `subs` and `outdir`, you will also need a default for `tag`. If not defined, this would most likely be `T1`.

bin/dm_mripicture.py
```diff
@@ -0,0 +1,78 @@
...
+        outpath = os.path.join(outdir, subject + '_T1.pdf')
+
+        # Output Image
+        plotting.plot_anat(imgpath, cut_coords=(-20, -10, 2), display_mode='x',
+                           annotate=False, draw_cross=False, vmin=100,
+                           vmax=1100, threshold='auto')
+        plt.savefig(outpath, bbox_inches='tight', pad_inches=0,
+                    dpi=1000, facecolor='w', edgecolor='k')
+
+    print('Saved all output to', outdir)
```
josephmje, line 74:
```suggestion
    print(f'Saved all output to {outdir}')
```

josephmje (approved):
Looks good to me @kyjimmy! Also, re the `cut_coords`, I didn't realize nilearn is using MNI coordinates. It might be fine to keep this hard-coded.

kyjimmy:
> Looks good to me @kyjimmy! Also, re the `cut_coords`, I didn't realize nilearn is using MNI coordinates. It might be fine to keep this hard-coded.

Thank you @josephmje. I just realized that I need to update the doc/requirements.txt and add nilearn to the list.

josephmje:
> I just realized that I need to update the doc/requirements.txt and add nilearn to the list.

You'll want to update the `setup.cfg` file instead actually with the nilearn version you're using. the `docs/requirements.txt` file only contains the minimal python packages required for building the datman documentation.

josephmje (approved):

jerdra (approved):

PR #305, opened 2021-06-07
[FIX] framewise displacement mis-calculation

jerdra:


bin/transfer_feenics_motion.py
```diff
@@ -143,6 +151,8 @@ def main():
     fmriprep = arguments["--fmriprep"] or cfg.get_path("fmriprep")
     subjects = arguments["--subject"]
 
+    fd_thres = arguments["--fd"] or 0.5
```
DESm1th, line 154: It might be good to document that the default is 0.5 in the docstring (unless this would be obvious to a knowledgeable, i.e. non-me, user)

DESm1th (approved):

bin/transfer_feenics_motion.py
jerdra: its the usual standard for FD in the lab, buut i agree about always being explicit!

DESm1th (approved):

DESm1th:
@josephmje @jerdra  Is it ok to merge this?

jerdra:
I think so!

PR #307, opened 2021-06-16
[ENH] Add dm_parse_faces

jerdra:
Fixed up PEP8 issues and added a bit more clarity to logging

### DIFF
--- a/bin/dm_parse_faces.py
+++ b/bin/dm_parse_faces.py
@@ -40,6 +40,7 @@ logging.basicConfig(level=logging.WARN,
                     format="[%(name)s] %(levelname)s: %(message)s")
 logger = logging.getLogger(os.path.basename(__file__))
 
+
 def read_eprime(eprimefile):
     '''
     Read in ePrime file with appropriate encoding
@@ -158,7 +159,7 @@ def main():
         sessions = [sessions]
 
     print_sessions = "\n".join(sessions)
-    logger.debug(f"Running FACES parser for session(s):")
+    logger.debug("Running FACES parser for session(s):")
     logger.debug(f"{print_sessions}")
 
     for ses in sessions:
--- a/bin/dm_parse_faces.py
+++ b/bin/dm_parse_faces.py
@@ -1,0 +1,251 @@
+#!/usr/bin/env python
+"""FACES Task Parser
+
+Converts eprime text file for FACES task into BIDS .tsv files.
+
+Usage:
+    dm_parse_faces.py [options] <study>
+    dm_parse_faces.py [options] <study> [<session>]
+
+Arguments:
+    <study>     Name of the study to process e.g. OPT
+    <session>   Datman name of session to process e.g. OPT01_UP1_UP10044_02_01
+
+Options:
+    --output-dir OUT_DIR    Specify an alternate output directory
+                            [Default:
+                            $DATMAN_PROJECTSDIR/{study}/data/nii/{session}]
+    --debug                 Set log level to debug
+    --dry-run               Perform a run but do not create outputs
+
+Outputs the trial number, onset time, accurracy, and reaction time values
+for each trial.
+"""
+
+from docopt import docopt
+import codecs
+import logging
+import pandas as pd
+import numpy as np
+import re
+import os
+import glob
+
+import datman.config
+import datman.scanid
+
+logging.basicConfig(level=logging.WARN,
+                    format="[%(name)s] %(levelname)s: %(message)s")
+logger = logging.getLogger(os.path.basename(__file__))
+
+_to_esc = re.compile(r'\s|()')
+
+
+def read_eprime(eprimefile):
+    '''
+    Read in ePrime file with appropriate encoding
+    '''
+    eprime = codecs.open(eprimefile, "r", encoding="utf-16", errors="strict")
+    lines = []
+    for line in eprime:
+        lines.append(str(line))
+    return lines
+
+
+def find_all_data(eprime, tag):
+    '''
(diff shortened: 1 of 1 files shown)


bin/dm_parse_faces.py
```diff
@@ -0,0 +1,251 @@
...
+        logger.info(f"Parsing {ses}...")
+
+        try:
+            ident = datman.scanid.parse(ses)
+        except datman.scanid.ParseException:
+            logger.error(f"Skipping task folder with malformed ID {ses}")
+            continue
+        ses_path = os.path.join(task_path, ses)
+
+        task_files = glob.glob(ses_path + '/*.txt')
```
kyjimmy, line 166: right now, all the nback task files are located in /RESOURCES/{subject}/behav/. Do I need to change the input directory or move the data to $ses_path?

jerdra: two things:

1. This should only affect the FACES task files, so your nback shouldn't be affected by this PR!
2. The way we handle tasks is to source it from the RESOURCES directory and rename + link to {study}/task/{session} - this is what `dm_task_files.py` does

bin/dm_parse_faces.py
```diff
@@ -0,0 +1,251 @@
...
+                    'trial_type':
+                    'Shapes' if 'Shape' in str(b) else 'Faces',
+                    'response_time':
+                    get_event_value(b, 'StimSlide.RT:'),
+                    'accuracy':
+                    get_event_value(b, 'StimSlide.ACC:'),
+                    'correct_response':
+                    map_response(get_event_value(b, 'CorrectResponse:')),
+                    'participant_response':
+                    map_response(get_event_value(b, 'StimSlide.RESP:'))
```
kyjimmy, line 216: "onset": 
                        get_event_value(b, "Stim.OnsetTime:"),
                        "duration": 
                        get_event_value(b, "Stim.OnsetToOnsetTime:"),
                        "trial_type": 
                        get_event_value(b, "BlockType:"),
                        "response_time": 
                        get_event_value(b, "Stim.RT:"),
                        "accuracy": 
                        get_event_value(b, "Stim.ACC:"),
                        "correct_response": 
                        map_response( get_event_value(b, "CorrectResponse:")),
                        "participant_response": 
                        map_response(get_event_value(b, "Stim.RESP:")),

kyjimmy: I do notice that our output text file is a little different from other projects.

kyjimmy (commented):
Variables name in the output text file of TAY project is different from other other projects.

bin/dm_parse_faces.py
kyjimmy: Instead of 'Shapes' and 'Faces', we have "0-Back" and "2-Back" as the trial type

jerdra: it sounds like we'll need a separate parser for TAY n-Back (or some refactoring to build a more general solution), but this, at the moment is for OPTIMUM FACES!

kyjimmy: yes, I felt like we were talking about different things. Sorry for the confusion.

Just one minor edit
`get_event_value(b, 'StimSlide.OnsetToOnsetTime:'),`

kyjimmy (approved):
Looks great. Thank you @jerdra for putting everything together! :)

bin/dm_parse_faces.py
```diff
@@ -0,0 +1,251 @@
...
+                })\
+                .astype({
+                    "correct_response": "Int64",
+                    "participant_response": "Int64",
+                    "accuracy": "Int64"
+                })
+
+            log_head, log_tail = os.path.split(eprimefile)
+
+            if not out_dir:
```
kyjimmy, line 236: `    --output-dir OUT_DIR    Specify an alternate output directory                            [Default:                            $DATMAN_PROJECTSDIR/{study}/data/nii/{session}]`

Wouldn't `out_dir` never be empty since there's a default value? Please correct me if I am wrong.

bin/dm_parse_faces.py
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/env python
+"""FACES Task Parser
+
+Converts eprime text file for FACES task into BIDS .tsv files.
+
+Usage:
+    dm_parse_faces.py [options] <study>
+    dm_parse_faces.py [options] <study> [<session>]
```
josephmje, line 8: Would this need to take multiple sessions? Can add `[<session>]...`

bin/dm_parse_faces.py
```diff
@@ -0,0 +1,251 @@
...
+                })
+
+            log_head, log_tail = os.path.split(eprimefile)
+
+            if not out_dir:
+                out_dir = os.path.join(
+                    nii_path, ident.get_full_subjectid_with_timepoint())
+
+            file_name = os.path.join(out_dir, f"{ses}_FACES.tsv")
+            os.makedirs(os.path.dirname(file_name), exist_ok=True)
```
josephmje, line 241: Should this bit be moved to within the `if not dryrun` statement?

josephmje (dismissed):
Thanks @jerdra! Just left some minor comments.

bin/dm_parse_faces.py
jerdra: `out_dir` would be empty if the `--output-dir` arg isn't specified. The default in the docstring is just so the user knows it'll be put into the configured archive folder, it won't actually set a value (the docstring parser will silently fail to make it a real value).

DESm1th (approved):

josephmje (approved):

PR #309, opened 2021-08-11
fix CBF data extraction issue

kyjimmy:
The CBF data may have "unknown" series description as the description's name is time-stamped, leading to unsuccessful scan.set_datman_name(str(ident), tags.series_map). Therefore, I created a new function to rename the series description based on series number, and it searches inside a dictionary:
scan_exception = {("unknown","1350"):"CBF"}
But I am not sure where to put this dictionary, maybe TAY_settings.yml?

In addition, CBF is a derived scan so I added another condition.


DESm1th:
Hey Jimmy! I like the idea behind this solution :) Yeah, I'd put such a setting inside tay_settings.yml. The only issue is there's nowhere in the code where scan_exception is being set/read currently. You may want to add an option to the 'process_scans' function and somewhere in dm_xnat_extract.py where the scan exceptions/series description override gets read from config

bin/dm_xnat_extract.py
```diff
@@ -430,6 +430,7 @@ def process_scans(xnat, ident, xnat_experiment):
 
     # load the export info from the site config files
     tags = cfg.get_tags(site=ident.site)
+    scan_exception = cfg.get_key('ScanException', site=ident.site)
```
kyjimmy, line 433: probably need exception handling?

josephmje:
@kyjimmy Sorry for the delay in getting to this. Is this PR still relevant after our email with Brad? I think the CBF image was just a poor skull-stripped version of the ASL scan.

kyjimmy:
Hi @josephmje! Right, I don't think it's relevant anymore. I will close the pull request. Thank you!

PR #317, opened 2021-11-16
Added inital xnat_to_bids function and error checking for it

slimnsour:
This is my base work for getting xnat_to_bids working on our system. This adds the function `xnat_to_bids` that pulls the data directly from xnat into a tmp folder, and then dcm2bids is run on that tmp folder with an appropriate json for that study. This is called instead of the usual xnat to nii pull if you call `dm_xnat_extract.py` with the tag `--use-dcm2bids`. Several arguments for dcm2bids are also allowed to be controlled (like `--dcm-config`, `--forceDcm2niix`, and `--clobber`). 

I would greatly appreciate feedback on this! Right now I have some design decisions that I'm unsure of as well:
 - The default dcm2bids json location for each study is at `/archive/data/$STUDY/metadata/dcm2bids.json`. I think this might be a good idea instead of forcing it into the yamls for each of the studies, since the jsons mapping each study could overwhelm the file contents. Instead, we can have a separate location (its own repo or a folder or in [datman-config](https://github.com/TIGRLab/datman-config)) for all the jsons and then soft link them to `metadata/dcm2bids.json`. An example folder is already in [our bids_dev repo](https://github.com/TIGRLab/bids_dev/tree/main/config).
 - The default bids folder is at `/archive/data/$STUDY/data/bids`. I believe this makes sense, especially with the option `--bids-out` that allows you to override it for testing purposes.
 - Dcm2Bids is currently being imported as a package and being called as a function. I did this as a configurable option since [our Dcm2Bids](https://github.com/TIGRLab/Dcm2Bids) is already a python package, but this would require datman to depend on Dcm2Bids now. I'm open to other options, namely using a singularity container. I already have the container with the newest dcm2niix and our dcm2bids installed, it's just a matter of switching it out if we want to use that instead.
 
I still have other features I might add in the future (fixing fmap json intendedFor's, defacing), but I'd like to get the base functionality out there to get your guys' opinions on it.

### DIFF
--- a/bin/dm_xnat_extract.py
+++ b/bin/dm_xnat_extract.py
@@ -84,7 +84,6 @@ import sys
 import re
 import zipfile
 
-from docopt import docopt
 import pydicom as dicom
 
 import datman.dashboard as dashboard
@@ -96,6 +95,8 @@ import datman.scanid
 import datman.exceptions
 
 from dcm2bids import Dcm2bids
+from pathlib import Path
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
 
 logger = logging.getLogger(os.path.basename(__file__))
 
@@ -108,12 +109,20 @@ DRYRUN = False
 db_ignore = False  # if True dont update the dashboard db
 wanted_tags = None
 
-keep_dcm = False
-dcm2bids_config = None
-use_dcm2bids = False
-bids_out = None
-force_dcm2niix = False
-clobber = False
+
+def _is_dir(path, parser):
+    """Ensure a given directory exists."""
+    if path is None or not Path(path).is_dir():
+        raise parser.error(f"Directory does not exist: <{path}>")
+    return Path(path).absolute()
+
+
+def _is_file(path, parser):
+    """Ensure a given directory exists."""
+    if path is None or not Path(path).is_file():
+        raise parser.error(f"Directory does not exist: <{path}>")
+    return Path(path).absolute()
+
 
 def main():
     global AUTH
@@ -123,26 +132,123 @@ def main():
     global wanted_tags
     global db_ignore
 
-    arguments = docopt(__doc__)
-    verbose = arguments['--verbose']
-    debug = arguments['--debug']
-    quiet = arguments['--quiet']
-    study = arguments['<study>']
-    experiment = arguments['<experiment>']
-    wanted_tags = arguments['--tag']
-    username = arguments['--username']
-    db_ignore = arguments['--dont-update-dashboard']
-    SERVER_OVERRIDE = arguments['--server']
+    parser = ArgumentParser(
+        description="Extracts data from XNAT archive folders into a "
+                    "few well-known formats.",
+        formatter_class=ArgumentDefaultsHelpFormatter,
+    )
 
-    # dcm2bids commands
-    use_dcm2bids = arguments['--use-dcm2bids']
-    keep_dcm = arguments['--keep-dcm']
-    dcm2bids_config = arguments['--dcm-config']
-    bids_out = arguments['--bids-out']
-    force_dcm2niix = arguments['--force-dcm2niix']
-    clobber = arguments['--clobber']
-    
-    if arguments['--dry-run']:
+    g_main = parser.add_argument_group(
+        "Options for choosing data from XNAT to extract"
+    )
+    g_main.add_argument(
+        "study",
+        action="store",
+        help="Nickname of the study to process",
+    )
+    g_main.add_argument(
+        "experiment",
+        action="store",
+        nargs="?",
+        help="Full ID of the experiment to process",
+    )
+    g_main.add_argument(
+        "--blacklist", action="store", metavar="FILE",
+        type=lambda x: Path(x).isfile(),
+        help="Table listing series to ignore override the "
+             "default metadata/blacklist.csv"
+    )
+    g_main.add_argument(
+        "--server", action="store", metavar="URL",
+        help="XNAT server to connect to, overrides the server "
+             "defined in the configuration files."
+    )
+    g_main.add_argument(
+        "-u", "--username", action="store", metavar="USER",
+        help="XNAT username. If specified then the environment "
+             "variables (or any credential files) are ignored "
+             "and you are prompted for a password. Note that if "
+             "multiple servers are configured for a study the "
+             "login used should be valid for all servers.."
+    )
+    g_main.add_argument(
+        "--dont-update-dashboard", action="store_true", default=False,
+        help="Dont update the dashboard database"
+    )
+    g_main.add_argument(
+        "-t",
+        "--tag",
+        action="store",
+        metavar="tag,...",
+        nargs="?",
+        help="List of scan tags to download"
+    )
+
+    g_dcm2bids = parser.add_argument_group(
+        "Options for using dcm2bids"
+    )
+    g_dcm2bids.add_argument(
+        "--use-dcm2bids", action="store_true", default=False,
+        help="Pull xnat data and convert to bids using dcm2bids"
+    )
+    g_dcm2bids.add_argument(
+        "--bids-out", action="store", metavar="DIR",
+        type=lambda x: _is_dir(x, parser),
+        help="Path to output bids folder"
+    )
+    g_dcm2bids.add_argument(
+        "--dcm-config", action="store", metavar="FILE",
+        type=lambda x: _is_file(x, parser),
+        help="Path to dcm2bids config file"
+    )
+    g_dcm2bids.add_argument(
+        "--keep-dcm", action="store_true", default=False,
+        help="Keep raw dcm pulled from xnat in temp folder"
+    )
+    g_dcm2bids.add_argument(
+        "--force-dcm2niix", action="store_true", default=False,
+        help="Force dcm2niix to be rerun in dcm2bids"
+    )
+    g_dcm2bids.add_argument(
+        "--clobber", action="store_true", default=False,
+        help="Clobber previous bids data"
+    )
+
+    g_perfm = parser.add_argument_group("Options for logging and debugging")
+    g_perfm.add_argument(
+        "-d", "--debug", action="store_true",
+        default=False,
+        help="Show debug messages"
+    )
+    g_perfm.add_argument(
+        "-q", "--quiet", action="store_true",
+        default=False,
+        help="Minimal logging"
+    )
+    g_perfm.add_argument(
+        "-v", "--verbose", action="store_true",
+        default=False,
+        help="Maximal logging"
+    )
+    g_perfm.add_argument(
+        "-n", "--dry-run", action="store_true",
+        default=False,
+        help="Do nothing"
+    )
+
+    args = parser.parse_args()
+    verbose = args.verbose
+    debug = args.debug
+    quiet = args.quiet
+    study = args.study
+    experiment = args.experiment
+    wanted_tags = args.tag
+    username = args.username
+    db_ignore = args.dont_update_dashboard
+    SERVER_OVERRIDE = args.server
+    use_dcm2bids = args.use_dcm2bids
+
+    if args.dry_run:
         DRYRUN = True
         db_ignore = True
 
@@ -162,11 +268,11 @@ def main():
 
     for xnat, project, experiment in experiments:
         if (use_dcm2bids):
-            dcm2bids_opt = Dcm2BidsConfig(keep_dcm=keep_dcm,
-                                          dcm2bids_config=dcm2bids_config,
-                                          bids_out=bids_out,
-                                          force_dcm2niix=force_dcm2niix,
-                                          clobber=clobber)
+            dcm2bids_opt = Dcm2BidsConfig(keep_dcm=args.keep_dcm,
+                                          dcm2bids_config=args.dcm_config,
+                                          bids_out=args.bids_out,
+                                          force_dcm2niix=args.force_dcm2niix,
+                                          clobber=args.clobber)
             xnat_to_bids(xnat, project, experiment, dcm2bids_opt)
         else:
             process_experiment(xnat, project, experiment)
@@ -312,7 +418,8 @@ def process_experiment(xnat, project, ident):
 
 
 class Dcm2BidsConfig(object):
-    def __init__(self, keep_dcm=False, bids_out=None, force_dcm2niix=False, clobber=False, dcm2bids_config=None):
+    def __init__(self, keep_dcm=False, bids_out=None,
+                 force_dcm2niix=False, clobber=False, dcm2bids_config=None):
         self.keep_dcm = keep_dcm
         self.force_dcm2niix = force_dcm2niix
         self.clobber = clobber
@@ -320,22 +427,14 @@ class Dcm2BidsConfig(object):
         self.bids_out = bids_out
         if dcm2bids_config is None:
             try:
-                self.dcm2bids_config = datman.utils.locate_metadata("dcm2bids.json", config=cfg)
+                self.dcm2bids_config = datman.utils.locate_metadata(
+                    "dcm2bids.json", config=cfg
+                )
             except FileNotFoundError:
                 logger.error("No config file available for study {}."
-                                "".format(study))
+                             "".format(cfg.study_name))
         if bids_out is None:
             self.bids_out = cfg.get_path("bids")
-
-    def run(self, input_dir, sub_id, ses_id=None, log_level="INFO"):
-        try:
-            dcm2bids_app = Dcm2bids(input_dir, sub_id, self.dcm2bids_config, output_dir=self.bids_out, session=ses_id, 
-                                    clobber=self.clobber, forceDcm2niix=self.force_dcm2niix, log_level=log_level)
-            dcm2bids_app.run()
-        except Exception as e:
-            logger.error("Dcm2Bids failed to run for experiment {}. "
-                            "{}: {}".format(experiment_label, type(e).__name__, e))
-            return
 
 
 def xnat_to_bids(xnat, project, ident, dcm2bids_opt):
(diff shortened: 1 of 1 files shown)


bin/dm_xnat_extract.py
```diff
@@ -139,7 +161,27 @@ def main():
...
+            if bids_out is None:
+                bids_out = cfg.get_path("bids")
+            xnat_to_bids(xnat, 
+                        project, 
+                        experiment,
+                        keep_dcm=keep_dcm,
+                        dcm2bids_config=dcm2bids_config,
+                        bids_out=bids_out,
+                        forceDcm2niix=forceDcm2niix,
+                        clobber=clobber)
```
jerdra, line 182: I'd refactor this and bundle these parameters as a dcm2bids Config object/dataclass

bin/dm_xnat_extract.py
```diff
@@ -139,7 +161,27 @@ def main():
...
+        else:
+            process_experiment(xnat, project, experiment)
```
jerdra, line 184: I'm even more convinced that we should use a sub-parser now! You could set up `subparser.set_defaults(function=<target_func>)` to simplify the logic:

https://stackoverflow.com/questions/49038616/argparse-subparsers-with-functions

bin/dm_xnat_extract.py
```diff
@@ -281,6 +323,45 @@ def process_experiment(xnat, project, ident):
...
+
+    try:
+        xnat_experiment = xnat.get_experiment(
+            project, ident.get_xnat_subject_id(), experiment_label)
+    except Exception as e:
+        logger.error("Unable to retrieve experiment {} from XNAT server. "
+                     "{}: {}".format(experiment_label, type(e).__name__, e))
+        return
+
+    # Make temporary folder for this subject
```
jerdra, line 343:
```suggestion
  
```
unnecessary comment

bin/dm_xnat_extract.py
```diff
@@ -281,6 +323,45 @@ def process_experiment(xnat, project, ident):
...
+    with datman.utils.make_temp_directory(prefix='xnat_to_bids_') as tempdir:
+        # Pull each of the scans from the subject's experiment
```
jerdra, line 345:
```suggestion
        
```
also not needed

bin/dm_xnat_extract.py
```diff
@@ -281,6 +323,45 @@ def process_experiment(xnat, project, ident):
...
+        for scan in xnat_experiment.scans:
+            scanTemp = get_dicom_archive_from_xnat(xnat, scan, tempdir)
+            if not scanTemp:
```
jerdra, line 348:
```suggestion
            scan_temp = get_dicom_archive_from_xnat(xnat, scan, tempdir)
            if not scan_temp:
```

bin/dm_xnat_extract.py
```diff
@@ -281,6 +323,45 @@ def process_experiment(xnat, project, ident):
...
+                logger.error("Failed getting series {} for experiment {} from XNAT"
+                            .format(scan.series, scan.experiment))
+                return
+        
+        # Once all the scans are downloaded, run dcm2bids on the pulled data
```
jerdra, line 353:
```suggestion
     
```

bin/dm_xnat_extract.py
```diff
@@ -281,6 +323,45 @@ def process_experiment(xnat, project, ident):
...
+        subDcmDir = os.path.join(tempdir, experiment_label, "scans")
+        try:
+            dcm2bids_app = Dcm2bids(subDcmDir, bids_sub, dcm2bids_config, output_dir=bids_out, session=bids_ses, 
+                                clobber=clobber, forceDcm2niix=forceDcm2niix, log_level="INFO")
```
jerdra, line 357:
```suggestion
        sub_dcm_dir = os.path.join(tempdir, experiment_label, "scans")
        try:
            dcm2bids_app = Dcm2bids(sub_dcm_dir, bids_sub, dcm2bids_config, output_dir=bids_out, session=bids_ses, 
                                clobber=clobber, forceDcm2niix=forceDcm2niix, log_level="INFO")
```

jerdra (changes requested):
Good work putting this together @slimnsour! Just have some comments on code-styling and 1 bigger ask to simplify the CLI interface. Lmk if you have questions :)

jerdra:
> I would greatly appreciate feedback on this! Right now I have some design decisions that I'm unsure of as well:
> 
> * The default dcm2bids json location for each study is at `/archive/data/$STUDY/metadata/dcm2bids.json`. I think this might be a good idea instead of forcing it into the yamls for each of the studies, since the jsons mapping each study could overwhelm the file contents. Instead, we can have a separate location (its own repo or a folder or in [datman-config](https://github.com/TIGRLab/datman-config)) for all the jsons and then soft link them to `metadata/dcm2bids.json`. An example folder is already in [our bids_dev repo](https://github.com/TIGRLab/bids_dev/tree/main/config).

Agreed, I think this is a good solution

> * The default bids folder is at `/archive/data/$STUDY/data/bids`. I believe this makes sense, especially with the option `--bids-out` that allows you to override it for testing purposes.

Perf! This is how `bidsify.py` currently works as well

> * Dcm2Bids is currently being imported as a package and being called as a function. I did this as a configurable option since [our Dcm2Bids](https://github.com/TIGRLab/Dcm2Bids) is already a python package, but this would require datman to depend on Dcm2Bids now. I'm open to other options, namely using a singularity container. I already have the container with the newest dcm2niix and our dcm2bids installed, it's just a matter of switching it out if we want to use that instead.

Maybe we could have datman have a `datman[bids]` setup.cfg option so that it becomes an optional dependency? This would require some refactoring (i.e moving dcm2bids import into a function, perhaps the one called by a subparser if that's what we end up going for)

bin/dm_xnat_extract.py
DESm1th: I second Jer's suggestion here

bin/dm_xnat_extract.py
DESm1th, line 357: Just to add on forceDcm2niix should probably also be force_dcm2niix

DESm1th (changes requested):
This looks good Salim, thank you! It looks like the 'making symlinks for the legacy datman folders' part was abandoned though. Without it a lot of our bin scripts and stuff will break at the moment so we might have to hold off on running nightly with the dcm2bids flag for a while

bin/dm_xnat_extract.py
```diff
@@ -139,7 +161,27 @@ def main():
         len(experiments), study))
 
     for xnat, project, experiment in experiments:
-        process_experiment(xnat, project, experiment)
+        if (use_dcm2bids):
+            # Ensure the config is available
+            if dcm2bids_config is None:
+                try:
+                    dcm2bids_config = datman.utils.locate_metadata("dcm2bids.json", config=cfg)
+                except datman.scanid.ParseException:
```
josephmje, line 169: Should this be a `FileNotFoundError` instead?

jerdra:
@slimnsour maybe instead of a sub-parser we could use `ArgumentParser.add_argument_group` like @josephmje did in #318?

slimnsour:
> @slimnsour maybe instead of a sub-parser we could use `ArgumentParser.add_argument_group` like @josephmje did in #318?
> 
> It would achieve separation of bids options from base options to mitigate overcrowding as we had originally wanted

Great suggestion! I've added that to the newest commit. 

For context, the subparsers in argparse (from my testing) requires that the option to select the subparser be the very first positional argument choose between the subparsers. So, if we wanted to have subparsers to separate between extracting using our usual datman way and dcm2bids, we would have to call it like `dm_xnat_extract.py {datman, dcm2bids} <study> ...`, which is _fine_, but it would break our current management scripts that assume `<study>` to be the first argument. If anyone can figure out a good way to do this then I would happy to implement it but it might be out of the scope of this PR.

Here's the new output of `dm_xnat_extract.py --help` using `argparse`:

```
usage: dm_xnat_extract.py [-h] [--blacklist FILE] [--server URL] [-u USER]
                          [--dont-update-dashboard] [-t [tag,...]]
                          [--use-dcm2bids] [--bids-out DIR]
                          [--dcm-config FILE] [--keep-dcm] [--force-dcm2niix]
                          [--clobber] [-d] [-q] [-v] [-n]
                          study [experiment]

Extracts data from XNAT archive folders into a few well-known formats.

optional arguments:
  -h, --help            show this help message and exit

Options for choosing data from XNAT to extract:
  study                 Nickname of the study to process
  experiment            Full ID of the experiment to process (default: None)
  --blacklist FILE      Table listing series to ignore override the default
                        metadata/blacklist.csv (default: None)
  --server URL          XNAT server to connect to, overrides the server
                        defined in the configuration files. (default: None)
  -u USER, --username USER
                        XNAT username. If specified then the environment
                        variables (or any credential files) are ignored and
                        you are prompted for a password. Note that if multiple
                        servers are configured for a study the login used
                        should be valid for all servers.. (default: None)
  --dont-update-dashboard
                        Dont update the dashboard database (default: False)
  -t [tag,...], --tag [tag,...]
                        List of scan tags to download (default: None)

Options for using dcm2bids:
  --use-dcm2bids        Pull xnat data and convert to bids using dcm2bids
                        (default: False)
  --bids-out DIR        Path to output bids folder (default: None)
  --dcm-config FILE     Path to dcm2bids config file (default: None)
  --keep-dcm            Keep raw dcm pulled from xnat in temp folder (default:
                        False)
  --force-dcm2niix      Force dcm2niix to be rerun in dcm2bids (default:
                        False)
  --clobber             Clobber previous bids data (default: False)

Options for logging and debugging:
  -d, --debug           Show debug messages (default: False)
  -q, --quiet           Minimal logging (default: False)
  -v, --verbose         Maximal logging (default: False)
  -n, --dry-run         Do nothing (default: False)
```

bin/dm_xnat_extract.py
slimnsour: Good catch! Fixed in f7486cb1a0cacf80a2f2ee0875159a7e5d5af3cb

bin/dm_xnat_extract.py
slimnsour: Thanks, fixed in f7486cb1a0cacf80a2f2ee0875159a7e5d5af3cb

slimnsour: Thanks, fixed in f7486cb1a0cacf80a2f2ee0875159a7e5d5af3cb

bin/dm_xnat_extract.py
slimnsour: Removed unnecessary comments in f7486cb1a0cacf80a2f2ee0875159a7e5d5af3cb

bin/dm_xnat_extract.py
slimnsour: Thanks for the catch, updated in f7486cb1a0cacf80a2f2ee0875159a7e5d5af3cb

bin/dm_xnat_extract.py
slimnsour: Removed unnecessary comments in f7486cb

bin/dm_xnat_extract.py
slimnsour: Removed unnecessary comments f7486cb

bin/dm_xnat_extract.py
```diff
@@ -26,6 +26,12 @@
                              login used should be valid for all servers.
     --dont-update-dashboard  Dont update the dashboard database
     -t --tag tag,...         List of scan tags to download
+    --use-dcm2bids           Pull xnat data and convert to bids using dcm2bids
```
josephmje, line 29: `argparse` should be able to build the usage arguments and options dynamically so the old docopt code in the docstring isn't as necessary.

bin/dm_xnat_extract.py
```diff
@@ -281,6 +323,45 @@ def process_experiment(xnat, project, ident):
...
+
+    # Get the identifying subject data
+    bids_sub = ident.get_bids_name()
+    bids_ses = ident.timepoint
+    experiment_label = ident.get_xnat_experiment_id()
+
+    try:
+        xnat_experiment = xnat.get_experiment(
+            project, ident.get_xnat_subject_id(), experiment_label)
+    except Exception as e:
```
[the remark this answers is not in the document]
slimnsour: Good catch! Fixed in f7486cb

bin/dm_xnat_extract.py
```diff
@@ -89,6 +94,10 @@
 import datman.scanid
 import datman.exceptions
 
+from dcm2bids import Dcm2bids
+from pathlib import Path
```
josephmje, line 98: This is nitpicky but we don't need both `pathlib` and `os.path`. Stick with one.

bin/dm_xnat_extract.py
```diff
@@ -26,6 +26,12 @@
                              login used should be valid for all servers.
     --dont-update-dashboard  Dont update the dashboard database
     -t --tag tag,...         List of scan tags to download
+    --use-dcm2bids               Pull xnat data and convert to bids using dcm2bids
+    --keep-dcm               Keep raw dcm pulled from xnat in temp folder
+    --dcm-config CONFIG      Path to dcm2bids config file
+    --bids-out BIDS_OUT      Path to output bids folder
+    --forceDcm2niix          Force dcm2niix to be rerun in dcm2bids
+    --clobber                Clobber previous bids data
```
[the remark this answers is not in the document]
slimnsour: Added the dataclass in the new commits! Thanks for the suggestion

bin/dm_xnat_extract.py
```diff
@@ -101,6 +110,20 @@
...
 
+def _is_dir(path, parser):
+    """Ensure a given directory exists."""
+    if path is None or not Path(path).is_dir():
+        raise parser.error(f"Directory does not exist: <{path}>")
+    return Path(path).absolute()
+
+
+def _is_file(path, parser):
+    """Ensure a given directory exists."""
```
josephmje, line 121:
```suggestion
    """Ensure a given file exists."""
```

bin/dm_xnat_extract.py
```diff
@@ -101,6 +110,20 @@
...
+    if path is None or not Path(path).is_file():
+        raise parser.error(f"Directory does not exist: <{path}>")
```
josephmje, line 123:
```suggestion
        raise parser.error(f"File does not exist: <{path}>")
```

bin/dm_xnat_extract.py
slimnsour: Added dataclass in the new commits, great suggestion!

bin/dm_xnat_extract.py
```diff
@@ -109,18 +132,123 @@ def main():
...
+                    "few well-known formats.",
+        formatter_class=ArgumentDefaultsHelpFormatter,
+    )
+
+    g_main = parser.add_argument_group(
+        "Options for choosing data from XNAT to extract"
+    )
+    g_main.add_argument(
+        "study",
+        action="store",
```
josephmje, line 146:
```suggestion
        action="store",
        required=True,
```

bin/dm_xnat_extract.py
slimnsour: Added the new argparse parser in the new commits but no subparser, see https://github.com/TIGRLab/datman/pull/317#issuecomment-974262017

bin/dm_xnat_extract.py
```diff
@@ -109,18 +132,123 @@ def main():
...
+    )
+    g_main.add_argument(
+        "experiment",
+        action="store",
+        nargs="?",
+        help="Full ID of the experiment to process",
+    )
+    g_main.add_argument(
+        "--blacklist", action="store", metavar="FILE",
+        type=lambda x: Path(x).isfile(),
```
josephmje, line 157:
```suggestion
        type=lambda x: _is_file(x, parser),
```

bin/dm_xnat_extract.py
```diff
@@ -109,18 +132,123 @@ def main():
...
+        nargs="?",
+        help="List of scan tags to download"
+    )
+
+    g_dcm2bids = parser.add_argument_group(
+        "Options for using dcm2bids"
+    )
+    g_dcm2bids.add_argument(
+        "--use-dcm2bids", action="store_true", default=False,
+        help="Pull xnat data and convert to bids using dcm2bids"
```
josephmje, line 192: Should the `--use-dcm2bids` option exist outside of this? And then, **only** if that option is used, the `g_dcm2bids` options are exposed.

bin/dm_xnat_extract.py
slimnsour: I committed this without realizing it will break: `TypeError: 'required' is an invalid argument for positionals`. The `required` flag doesn't work for positionals in `argparse` since they're already assumed to be inputted. I make the `experiment` argument optional by using `nargs` (which I now realize should be limited to 1 argument so that will be changed!)

bin/dm_xnat_extract.py
slimnsour: Good catch, the usage argments docstring was removed in c130ceeb5704aa3854f29b7ed461d167555a6744

bin/dm_xnat_extract.py
slimnsour: Good point, I switched it to only use `os.path` in c130ceeb5704aa3854f29b7ed461d167555a6744

bin/dm_xnat_extract.py
slimnsour: I moved it to the main group cause it makes more sense there like you said, but Im not sure how to only expose dcm2bids arguments when we read from the parser with `argparse`. Reading from a specific group is not supported and subparsers would break the usage I mentioned in https://github.com/TIGRLab/datman/pull/317#issuecomment-974262017

bin/dm_xnat_extract.py
josephmje: My mistake!

bin/dm_xnat_extract.py
jerdra: i think @josephmje wants to warn/prevent users from using BIDS related flags when --use-dcm2bids isn't selected. What we could do is to add a bit of code after parsing to check if any of the BIDS config exists when --dcm2bids *is not selected* and either:
1. Throw up a warning indicated that those args will be ignored
2. Use `ArgumentParser.error` with a message
3. 
I prefer option 2 personally as it'll prevent mistakes from running through

josephmje: Yup! Could do something like

```
if not args.use_dcm2bids and (args.bids_out or args.dcm_config or ...):
    parser.error("dcm2bids configuration required --use-dcm2bids")
```

slimnsour: Ah gotcha, thank you guys for laying that out! I added it in b1cb96bb5d86be38a4ca608afbec8ca4b629b15a

jerdra (approved):
Good work and thanks for incorporating our feedback @slimnsour!

josephmje (approved):

PR #318, opened 2021-11-17
ENH: Add defacing script

josephmje:
Added a script to perform defacing.

I'm looking for some suggestions on better implementation with datman and the dashboard:
- Should `--suffix-id` come from the study config?
- I ended up saving the pydeface command to the json file. Not sure if we should overwrite the original scan. My initial preference is to keep both files. It could be good to use both to create a mask overlay QC image and then if we want, we can replace the original image after QC

### DIFF
--- a/bin/dm_deface.py
+++ b/bin/dm_deface.py
@@ -1,0 +1,113 @@
+#!/usr/bin/env python
+"""
+Defaces anatomical data in the BIDS folder.
+
+Usage:
+    dm_deface.py [options] <study> [--bids-filter-suffix <bids_suffix>]... [--skip-bids-validation]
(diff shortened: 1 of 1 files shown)


DESm1th (dismissed):
Looks good Mike! Thank you. Is there any possibility of the deface command failing silently though?

jerdra:
> Looks good Mike! Thank you. Is there any possibility of the deface command failing silently though?

might make sense to call `pydeface.utils.deface_image` directly? Our inputs seem simple enough and we don't have to deal with weird cmdline arg construction or sub-process handling

josephmje:
> might make sense to call `pydeface.utils.deface_image` directly? Our inputs seem simple enough and we don't have to deal with weird cmdline arg construction or sub-process handling

Thanks for the suggestion! Done in b0a76c6

jerdra (approved):

slimnsour (approved):
Looks good to me! Just tested it out and it works as expected too

PR #320, opened 2022-01-06
Updates and bug fixes for dm_parse_faces

benselby:
Before the break I realized that there was a bug in the FACES parser that was resulting in output tsvs not being saved in the correct nii/<SUBJECT> directories. Also a few of the OPT session .txt files had UTF-16 encoding issues wherein the files were not being decoded properly due to a missing BOM, so I added some code to try specifying endianness which resolves most issues.

### DIFF
--- a/bin/dm_parse_faces.py
+++ b/bin/dm_parse_faces.py
@@ -139,7 +139,7 @@ def main():
     dryrun = arguments["--dry-run"]
 
     if verbose:
-        logger.seLevel(logging.INFO)
+        logger.setLevel(logging.INFO)
     if debug:
         logger.setLevel(logging.DEBUG)
     if quiet:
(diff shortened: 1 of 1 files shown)


DESm1th (approved):
Thanks for fixing this Ben!

bin/dm_parse_faces.py
```diff
@@ -45,11 +45,23 @@ def read_eprime(eprimefile):
...
+                lines.append(str(line))
+            if lines and 'Header Start' in lines[0]:
+                return lines
+
+        except UnicodeError as e:
+            logging.info(f"Failed to read {eprimefile} "
+                         f"with {enc} encoding: {e}")
+            continue
+
+    raise UnicodeError("Unable to find appropriate encoding")
```
jerdra, line 64: would an `IOError` be more appropriate here? A missing Header line doesn't have to mean its a encoding issue, could just be a corrupt file w/correct encoding. 

I'm nitpicking here, so feel free to ignore if you disagree

jerdra (approved):
Thanks for the fix!

PR #322, opened 2022-02-09
FIX: remove study invalid attribute access

jerdra:


josephmje (approved):

DESm1th (approved):
Thanks for catching this Jer, sorry about that!

PR #324, opened 2022-02-15
[FIX] Update circle ci ubuntu + python versions

DESm1th:
16.04's image is being phased out. This updates our config to use 20.04 instead and updates the python version to 3.9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
   build:
     working_directory: /tmp/src/datman
     machine:
-      image: ubuntu-1604:201903-01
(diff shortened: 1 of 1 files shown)


jerdra:
Thanks for the update @DESm1th, is there a reason to use 3.9.7/3.9.0 vs. just using one of the two?

DESm1th:
Not really! I just didn't want to pin the docker python version to a specific minor version just because the ubuntu image used that one. I can update it though if you'd prefer

jerdra:
Ahh ok ok it's not actually important then, works for me!

josephmje (approved):

PR #326, opened 2022-03-03
[FIX] Docs build failures

DESm1th:
Have to explicitly pin docutils below version 0.18.0 or the circleci build_docs step fails


josephmje (approved):

josephmje:
Thanks for this fix @DESm1th! I was wondering what do you think about migrating our CI to GitHub Actions?

DESm1th:
I'm not sure if it would be worth it yet. It's not as mature or featureful as circleci yet, and does some weird things with caching that would probably make our jobs slower atm.

PR #327, opened 2022-03-08
FIX: Fix/ea handle quit code

jerdra:
Ran into a special case where there was a "Quit" code at the end of the EA task file with `nan` entries causing a bunch of type errors. Fix is to remove the index associated with "Quit".

1. Refactor to remove code duplication: ce60dc3a05db6d51d551c56012018ac9c2d07fdf
2. Add check to handle quit signal: c46048b43d255624acf82eeb5ffd1b293f503f77, 407a06d68a6701ec7d23b135ade6b7a491e667c1
3. Apply code formatting and display error message on exception: 1e81ac4166f5e441d26e35c9ba59876a3e192684

### DIFF
--- a/bin/dm_parse_ea.py
+++ b/bin/dm_parse_ea.py
@@ -111,6 +111,12 @@ def clean_logfile(log_file):
         logger.warning(f"Removed {indices_dropped} rating registered "
                        "followed scanner responses")
 
+    last_entry = log_file.loc[log_file.shape[0] - 1]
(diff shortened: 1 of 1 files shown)


jerdra:
@josephmje @benselby could i get eyes on this? Would be nice to avoid calling my specific version of the EA script :)

DESm1th (approved):
Thanks for the fix and cleaning up the indentation, Jer!

PR #328, opened 2022-03-17
ENH: Update configuration options for defacing

josephmje:
This updates the behaviour of `dm_deface.py` to replace the original scan by default. If the `--separate` flag is used, the script will output a new file in the BIDS folder using the `acq-defaced` tag (can be used for SPN20 and RMA01 where we still need the original files for TMS).


DESm1th (approved):
Thanks for doing this Mike!

PR #330, opened 2022-04-20
[FIX] Fixed xnat_to_bids ignoring --clobber before pulling data

slimnsour:
Small fix with behavior of `xnat_to_bids`: Before it would download the zip from xnat before checking if the subject was already converted to bids in the destination folder, which would add unnecessary time to batch runs. Now it will check the destination folder for the subject before attempting to pull their data (unless the --clobber tag is used of course)

I'm also working on a doc page to illustrate how to use `xnat_to_bids`, but with this fix we can safely put `dm_xnat_extract.py $STUDY --use-dcm2bids` in nightly scripts as it will skip over converted subjects properly.

### DIFF
--- a/bin/dm_xnat_extract.py
+++ b/bin/dm_xnat_extract.py
@@ -433,6 +433,16 @@ def xnat_to_bids(xnat, project, ident, dcm2bids_opt):
                      "{}: {}".format(experiment_label, type(e).__name__, e))
         return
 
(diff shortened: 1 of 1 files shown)


DESm1th (approved):

DESm1th (approved):
Thanks for the quick fix Salim!

jerdra (approved):
good stuff, thanks for PR!

slimnsour:
Thanks guys!

PR #334, opened 2022-08-19
[ENH] Refactor dm_xnat_extract + allow nii folders to be replace with sym links to bids folders

DESm1th:
Ok this is a really big one (sorry guys), but most of it is just reshuffled code. The really important bits are the contents of `datman/exporters.py` which contains a set of classes to export to different formats. All the old `export_x_command` functions from xnat_extract are now classes in that file, as is the contents of the xnat_to_bids function, and I've pulled all the database updating code (which was spread through xnat_extract) into its own exporter. I've added a NiiLinkExporter here that takes a subject's bids folder and makes datman-style links for it in the nii folder.

Within xnat_extract itself the most important bit is the `export_scans` function which does the work of creating the exporters, downloading any raw data needed to run them, and then running them.

The changes in `datman/xnat` is code I've moved from dm_xnat_extract, and the changes elsewhere were small fixes for bugs I encountered while refactoring.


datman/exporters.py
```diff
@@ -0,0 +1,1093 @@
...
+        """
+        try:
+            with open(dest, "w") as output:
+                output.write(messages)
+        except Exception as exc:
+            logger.error(f"Failed writing dcm2niix errors to {dest}. "
+                         f"Reason - {type(exc).__name__} {exc} ")
+
+
+class NrrdExporter(SeriesExporter):
```
josephmje, line 980: I think we can get rid of the nrrd and mnc exporters. We're no longer using CIVET, which requires the mnc files. And our tractography pipelines have a nifti to nrrd converter built in.

DESm1th: Fair enough! I was just trying not to drop functionality but I can purge those classes :)

DESm1th:
Is it alright for me to merge this?

jerdra:
@DESm1th sorry this dropped off my radar! I'll review the last bits today after our meet!

datman/exporters.py
```diff
@@ -0,0 +1,1058 @@
...
+                continue
+
+            bids_files = sorted(
+                matches,
+                key=lambda x: int(parse_bids_filename(x).run)
+            )
+
+            dm_files = sorted(
+                dm_names[tag],
+                key=lambda x: int(parse_filename(x)[2])
```
jerdra, line 321: Could this be updated to be a more obvious what [2] means?

datman/exporters.py
```diff
@@ -0,0 +1,1058 @@
...
+        """
+        bids_niftis = []
+        for path, _, files in os.walk(self.bids_path):
+            niftis = filter_niftis(files)
+            for item in niftis:
+                basename = item.replace(get_extension(item), "")
+                bids_niftis.append(os.path.join(path, basename))
+        return bids_niftis
+
+    def match_dm_to_bids(self, dm_names, bids_names):
```
jerdra, line 289: Is this ever meant to be called outside the object? Would `dm_names` and `bids_names` ever be provided outside of using `self`?

datman/exporters.py
```diff
@@ -0,0 +1,1058 @@
...
+
+            dm_files = sorted(
+                dm_names[tag],
+                key=lambda x: int(parse_filename(x)[2])
+            )
+
+            for idx, item in enumerate(dm_files):
+                if idx >= len(bids_files):
+                    continue
+                name_map[item] = bids_files[idx]
```
jerdra, line 327: Should we log cases in which we have a mismatch between BIDS files and Datman files?

datman/exporters.py
```diff
@@ -0,0 +1,1058 @@
...
+            full_path = os.path.join(self.output_dir, dm_name + self.ext)
+            if not os.path.exists(full_path):
+                return False
+        return True
+
+    def needs_raw_data(self):
+        return False
+
+    def export(self, *args, **kwargs):
+        # Re run this before exporting, in case new BIDS files exist.
```
jerdra, line 489: A little bit confused about why this needs to be re-initialized. Is it expected that the state of the BIDS directory may change between initializing this class and calling its `export` function? Doesn't it kind of make the state variables of the class useless/misleading?

jerdra (approved):
Thanks for the massive re-factor Dawn! Nothing in my review should block a merge, more just questions for clarity since there're some bits where I can't fully understand the intent immediately!

datman/exporters.py
DESm1th: Unfortunately, I can't see a way to refactor it without writing a wrapper function to more explicitly unpack the result of parse_filename. I think I might add that in a follow up PR (along with any other code clarification changes from your other comments)

datman/exporters.py
DESm1th: Likely not, but I think I left it unprotected since I ended up calling it a lot while testing that the matching worked as expected. I can change this

datman/exporters.py
DESm1th: I originally had a log statement here, but it only ever caught files that datman doesn't care about/would never have existed in the datman folder anyway. It also ended up a bit spammy. I can re-add it though if you like!

datman/exporters.py
DESm1th: This was the result of an edge case when the data is exported for the very first time. The class would be initialized, the names would be mapped to an empty bids folder (so no nii files) and then when the exporters actually ran, on the first run through no links would be produced.

You're right that it might make more sense to remove the state variables for this reason, but I think there was a good reason why I didnt at the time. I'll have to look at it more closely again when I fix the issues you pointed out to try to remember why I kept them

datman/exporters.py
jerdra: Agreed re: future PR, i think maybe parse_filename return a namedtuple vs. a list could maybe add a lot of clarity to how it's being used moving forward

datman/exporters.py
jerdra: I'll leave that up to you! My preference would be to remove the extra arguments for clarity :see_no_evil:

datman/exporters.py
jerdra: Oh ok i took a second look at `dm_xnat_extract` and understand what you mean here now re: the edge case. I think you end up using the initialized state to check if the outputs exist right? It looks like you might still need it in that case!

PR #342, opened 2022-11-24
ENH: Add --match-series option to enforce that matches Series# is found

jerdra:
This appends the Series number onto the task files so that .tsv outputs can be properly associated with their scans. Currently the default behavior means that files will overwrite each other in an indeterminate manner if there are multiple files for the same EA part. 

@DESm1th i think maybe with blacklisting it might be a good idea to also remove .tsv files?


DESm1th (approved):

DESm1th:
That's a good point! I'll push a PR shortly to fix that (and another to fix the failing github actions). Thanks Jer!

slimnsour (approved):
Great addition Jer, and implementation looks good to me!

PR #344, opened 2022-11-28
[ENH] un-hardcode dm_blacklist_rm's list of directories to purge from

DESm1th:
This updates dm_blacklist_rm to use a config value (BlacklistRm) or command line arg (--path) to override the default list of paths to purge blacklisted files from.


jerdra (approved):

slimnsour (approved):
Great implementation Dawn!

PR #345, opened 2022-12-12
ENH: Enable XnatType key in ExportInfo

jerdra:
This enables users to use the `XnatType` key in lieu of `SeriesDescription` within a study's `ExportInfo`. The result is that tag identification will happen using XNAT's `Type` data field.

The justification behind using this feature is that XNAT's `Type` data-field is editable whereas SeriesDescription is not. As a result XNAT Project owners/maintainers can override the SeriesDescription if the original SeriesDescription doesn't properly allow us to identify an appropriate Tag.

Also this is a my bad moment but I habitually run yapf when saving code and it formatted a bunch of irrelevant lines.... For the actual changes made look at:

1. The chunk at L1565 in xnat.py
2. The chunk at L187 in tests/test_xnat.py

### DIFF
--- a/datman/xnat.py
+++ b/datman/xnat.py
@@ -23,10 +23,8 @@ logger = logging.getLogger(__name__)
 
 def get_server(config=None, url=None, port=None):
     if not config and not url:
-        raise XnatException(
-            "Can't construct a valid server URL without a "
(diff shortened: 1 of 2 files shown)


DESm1th (approved):
Thanks for doing this Jer! :)

PR #346, opened 2023-01-06
[FIX] Stop nightly run log issues

DESm1th:
- 5cfb6a0c6f13be72fb0b8baa22c210a4c6360bbe Allow the 'dir' attribute to be set for func bids file names (CLZ uses dir-AP/PA with rest)
- b3fc9c0bfc6591d68ed606c741bd7b52243b575b Stop a filename parsing error from crashing export for an entire session. 
- de5f811efc0e82d7687ecfd158aa73b7e69e30f6 Catch the correct exception when no XnatPort has been provided

### DIFF
--- a/datman/exporters.py
+++ b/datman/exporters.py
@@ -311,15 +311,28 @@ class NiiLinkExporter(SessionExporter):
                     name_map)
                 continue
 
-            bids_files = sorted(
-                matches,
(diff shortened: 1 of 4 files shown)


jerdra (approved):
thanks for these fixes, lgtm!

PR #375, opened 2025-08-28
[FIX] XNAT import: replace unsupported prearchive=false with Direct-Archive=true

rickyltwong:
### Description

Fixes DICOM upload to truly bypass prearchive by using the documented flag.

### Change

```
- "&prearchive=false&Ignore-Unparsable=true&inbody=true"
+ "&Direct-Archive=true&Ignore-Unparsable=true&inbody=true"
```

### Impact

Uploads go direct-to-archive.

### Compatibility

Requires XNAT 1.8.3+.

### Reference 

https://wiki.xnat.org/xnat-api/image-session-import-service-api

### DIFF
--- a/datman/xnat.py
+++ b/datman/xnat.py
@@ -706,7 +706,7 @@ class XNAT:
         upload_url = (
             f"{self.server}/data/services/import?project={project}"
             f"&subject={subject}&session={experiment}&overwrite=delete"
-            "&prearchive=false&Ignore-Unparsable=true&inbody=true")
+            "&Direct-Archive=true&Ignore-Unparsable=true&inbody=true")
 
         try:
             with open(filename, "rb") as data:


DESm1th:
I had actually tried this change myself months ago and found that it dramatically slowed things at the time (but this may have been before the server was upgraded). So I assumed that the now-undocumented option still worked as it used to and the new one wasn't truly bypassing the prearchive. Did you find that it actually sped things up? I'll experiment a bit

rickyltwong:
Direct-Archive option works as documented (with below conditions); I can't find any official reference to prearchive=false. The slowdown you observed is most likely due to the upload/workflow pattern, not the flag itself.

From my experience, if you pre-create an MR session (even an empty one) before uploading, XNAT sees the session as already archived and will fall back to the prearchive to perform the merge. That fallback is expected and is what causes the extra IO /slowdown. You might want to validate by running an upload of a clean new session with Direct-Archive=true (no pre-creation of the session), it should stay direct-to-archive and be fast.

- For true DA, I think we shouldn't pre-create the session, let the import create it then attach non-DICOM resources (technotes, niftis etc.) afterward. 
- For repair/merge jobs (where sessions already exist), use standard prearchive->review & merge-> archive flow with `overwrite=delete&overwrite_files=true`, **or** delete the broken session first and then DA a fresh upload.

Three more pre-requisites:
- Requires XNAT 1.8.3+.
- In my tests I only saw DA reliably work when an SCP receiver with DA enabled was available; when present you can also see the Direct Archive table. The docs say this isn’t strictly necessary, but my experience suggests it helps.
- Ensure the project’s Prearchive Settings allow overwriting existing files (so merges behave as expected).

DESm1th:
The `prearchive=false` is from older versions of the api (I think it existed for every version <=1.7). When the server issues first started (but before the upgrade this year to the current version) I tested without either flag, with `prearchive=false`, and with `Direct-Archive=true` and at the time the direct archive option was as slow as not using either option and also, as you said, often fell back to the prearchive anyway because of the resources files. So I assumed `prearchive=false` was still supported even though it's no longer documented, since it did appear to produce a speed increase (and actually made it directly into the archive reliably).

I'll re-verify with some test uploads over the weekend though!

rickyltwong:
I strongly doubt that the direct-archive option is only applicable when an SCP Receiver has direct-archive enabled.

image

It’s only after this setting is enabled that I can see the incoming session appear in the direct-archive table instead of the main pre-archive list on the pre-archive page. Otherwise, the direct-archive option is simply ignored.

image

Most importantly, the raw files are imported directly into the archive folder in the filesystem, rather than the prearchive folder. 

image

image


This greatly reduces file I/O, just as the official doc suggests.

Edited: I will find a chance to explore the prearchive option as I didn't get a chance to test with old versions of XNAT.

DESm1th:
Ah ok! Maybe it didn't work previously because the scp receiver didn't have it enabled. Alright I'll just switch us over to using `Direct-Archive=true`. It looks like xnat is phasing out `prearchive=false` (hence the lack of documentation for 1.8+) so it's probably not worth too much time investigating as long as the new option actually works.

